PR #21, opened 2016-12-29
Faster R-CNN (WIP)

fmassa:
This WIP PR implements Faster R-CNN based on https://github.com/rbgirshick/py-faster-rcnn .
There are lots of copy-paste from Girshick's repo, and a few files were copied as is.
I decided to fuse a number of operations together, but I'm not sure anymore it's the best way to go.

A few comments:

* There are two main classes: `FasterRCNN` and `RPN`. RPN implements the region proposal network, and faster r-cnn wraps almost all the detection logic (which might hurt flexibility at some point).
* Both classes behave differently if a ground-truth is provided, avoiding to have to write different network definitions for train/test.
* I tried to avoid a global `cfg` dict containing all the parameters, and instead pass the parameters as the constructors of the classes. But different train/test parameters still need to be defined in the model.
* There is currently no handling on the image scaling (originally present in `im_info[2]`), necessary for properly pruning small boxes in the RPN.
* I'll add optimized implementations for ROIPooling/nms/etc later on using cffi.

I'm opening this PR to get some feed-back on the general structure of the code.
Overall, I'm starting to think that it might be better to use the same structure as the one in py-faster-rcnn from Girshick repo.

Code cleanup and organization is required.

### DIFF
--- a/fast_rcnn/README.md
+++ b/fast_rcnn/README.md
@@ -6,8 +6,9 @@ python main.py PATH_TO_DATASET
 
 ## Things to add/change/consider
 * where to handle the image scaling. Need to scale the annotations, and also RPN filters the minimum size wrt the original image size, and not the scaled image
+* should image scaling be handled in FasterRCNN class?
 * properly supporting flipping
 * best way to handle different parameters in RPN/FRCNN for train/eval modes
-* uniformize Variables, they should be provided by the user and not processed by me
-* should image scaling be handled in FasterRCNN class?
+* uniformize Variables, they should be provided by the user and not processed by FasterRCNN/RPN classes
 * general code cleanup, lots of torch/numpy mixture
+* should I use a general config file?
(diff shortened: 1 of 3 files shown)


bhack:
Have you tested https://github.com/longcw/faster_rcnn_pytorch?
/cc @longcw

fmassa:
@bhack I haven't, and I didn't have the time to finish this properly.
Given that there are already a number of pytorch implementations of object detection algorithms in pytorch, I'll close this one for the time being.
If I find some time to finish this up with a simple interface, I'll send a new PR.

bhack:
@KaimingHe In the plan of releasing Mask rcnn there will be also a faster-rnn pytorch implementation merged in this repository?

bhack:
A TF WIP Mask r-cnn effort was started at https://github.com/CharlesShang/FastMaskRCNN. Actually there is still no public reference implementation of the paper so we will see what kind of accurancy can be reproduced.

PR #22, opened 2017-01-01
Residual Networks

szagoruyko:
Still improving the doc and README.

### DIFF
--- a/resnet/main.py
+++ b/resnet/main.py
@@ -230,7 +230,7 @@ def resnet(depth, width, num_classes, dropout):
         y = F.conv2d(o1, params[base+'.conv0'], stride=stride, padding=1)
         o2 = activation(y, params, stats, base+'.bn1', mode)
         if opt.dropout > 0:
-            o2 = F.dropout(o2, p=0.5, training=mode)
+            o2 = F.dropout(o2, p=dropout, training=mode)
         z = F.conv2d(o2, params[base+'.conv1'], stride=1, padding=1)
         if base + '.convdim' in params:
             return z + F.conv2d(o1, params[base+'.convdim'], stride=stride)


resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+parser.add_argument('--ngpu', default=1, type=int,
+                    help='number of GPUs to use for training, 0 to for CPU')
+parser.add_argument('--gpu_id', default='0', type=str,
+                    help='id(s) for CUDA_VISIBLE_DEVICES')
+
+# TODO: move this to main
+opt = parser.parse_args()
+print 'parsed options:', vars(opt)
+
+os.environ['CUDA_VISIBLE_DEVICES'] = opt.gpu_id
```
edgarriba: @szagoruyko you could use here `torch.cuda.set_device(opt.gpu_id)`

apaszke: You should never use `set_device` in regular scripts, we've added it only for iPython notebooks where it's sometimes convenient. `CUDA_VISIBLE_DEVICES` is better, because it will prevent any memory alocation on unused GPUs (every PyTorch process uses ~250MB memory on each GPU it sees).

resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+parser.add_argument('--dtype', default='float', type=str)
+parser.add_argument('--nthread', default=4, type=int)
+parser.add_argument('--dropout', default=0, type=float)
+
+# Training options
+parser.add_argument('--batchSize', default=128, type=int)
+parser.add_argument('--lr', default=0.1, type=float)
+parser.add_argument('--epochs', default=200, type=int, metavar='N',
+                    help='number of total epochs to run')
+parser.add_argument('--weightDecay', default=0.0005, type=float)
```
apaszke: Can you make that `weight_decay`, so it's consistent with other params?

resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+
+# Device options
+parser.add_argument('--save', default='/tmp/cifar', type=str,
+                    help='save parameters and logs in this folder')
+parser.add_argument('--ngpu', default=1, type=int,
+                    help='number of GPUs to use for training, 0 to for CPU')
+parser.add_argument('--gpu_id', default='0', type=str,
+                    help='id(s) for CUDA_VISIBLE_DEVICES')
+
+# TODO: move this to main
```
apaszke: ?

resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+        self.pad = pad
+
+    def __call__(self, tensor):
+        padded = self.module.forward(tensor.unsqueeze(0))[0]
+        x = np.random.randint(0, self.pad*2)
+        y = np.random.randint(0, self.pad*2)
+        return padded[:, y:y+tensor.size(1), x:x+tensor.size(2)].contiguous()
+
+
+def get_iterator(dataset, mode):
```
apaszke: How about caling `mode` `train`? That would be more readable

resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+    else:
+        transforms = [normalize, totensor]
+
+    dataset_constr = getattr(datasets, dataset)
+    ds = dataset_constr('.', train=mode, download=True)
+    data = ds.train_data if mode else ds.test_data
+    labels = ds.train_labels if mode else ds.test_labels
+    ds_tnt = tnt.dataset.TensorDataset([data.transpose(0, 2, 3, 1), labels])
+    ds_tnt = ds_tnt.transform({0: tnt.transform.compose(transforms)})
+    return ds_tnt.parallel(batch_size=opt.batchSize, shuffle=mode,
```
apaszke: Can't you use core data loaders in here? I can't see any functionality that's not supported (TensorDataset is also in the core).

resnet/main.py
```diff
@@ -0,0 +1,400 @@
...
+    :param depth: total number of layers
+    :param width: multiplier for number of feature planes in each convolution
+    :param num_classes: number of output neurons in the top linear layer
+    :return:
+      f: function that defines the model
+      params: optimizable parameters dict
+      stats: batch normalization moments
+    """
+    assert (depth - 4) % 6 == 0, 'depth should be 6n+4'
+    n = int((depth - 4) / 6)
```
apaszke: Use `//` and remove `int`.

resnet/main.py
edgarriba: I see. Does it make sense to include it to the core like this? At least for me having this interface is more intuitive to set the device.

```python
def set_device(id):
   os.environ['CUDA_VISIBLE_DEVICES'] = opt.gpu_id

# call it from core instead of cuda since I guess in the future you
# won't support only cuda devices
torch.set_device(whatever_id)
```

jekbradbury: huh, even without CUDA_VISIBLE_DEVICES I never get 250MB memory on non-active GPUs (unlike with Lua Torch)

szagoruyko:
will make a separate repo

taion:
@szagoruyko Was this version of the code released anywhere? The top hit I can see on Google for Wide ResNets in PyTorch is https://github.com/xternalz/WideResNet-pytorch, which looks to be different.

szagoruyko:
@taion no, never released anywhere. you can find the latest version of this code here https://gist.github.com/szagoruyko/f1c70fef94daa4cea4bff497dc9c9a7b

szagoruyko:
finally released the code here: https://github.com/szagoruyko/wide-residual-networks/tree/master/pytorch

PR #37, opened 2017-01-18
make PTB language model example use torchtext

jekbradbury:
Also arbitrarily changed the seed so that it still gives perplexity 113


siddsach:
This commit makes sense! I haven't found a single good example of usage of torchtext in a sequence prediction problem. Why wasn't it merged into the example?

PR #40, opened 2017-01-18
Add logistic regression example

lukeyeager:
When learning how to use pytorch, I found it useful to write this example. I tried to rely on low-level operations rather than high-level abstractions to make it clear what's going on.

Feel free to close this - it's probably too basic for an example. I just thought I'd make the PR so others can find it in case it's useful.


soumith:
thanks :)

PR #44, opened 2017-01-18
Make mnist_hogwild python 2.7 compatible

smartkiwi:
multiprocessing in python 2.7 doesn't support multiprocessing.set_start_method method - so we'll skip it


colesbury:
I think we should just delete this line. It was part of a CUDA + hogwild example and isn't necessary here.

soumith:
thanks!

PR #45, opened 2017-01-19
add requirements

crcrpar:
Hi!

I followed examples/vae/README.md
```
pip install -r requirements.txt
python main.py
```
but got the error below
> ImportError: No module named 'PIL'

so I added `pillow` to requirements.txt

### DIFF
--- a/vae/requirements.txt
+++ b/vae/requirements.txt
@@ -1,3 +1,4 @@
 torch
 tqdm
 six
+pillow


colesbury:
Hi @crcrpar, thanks for catching this. I think it's actually missing the [`torchvision`](https://github.com/pytorch/vision) package as a dependency, which in turn depends on pillow.

soumith:
thank you! fixed it correctly to `torchvision` rather than `pillow` via https://github.com/pytorch/examples/commit/363a9470d60479bce69664d101f02cda6ddf16a9

crcrpar:
@colesbury @soumith 
understood.

PR #47, opened 2017-01-19
Triplet Network [WIP]

edgarriba:
This WIP PR implements a triplet network based on **TFeat shallow convolutional patch descriptor** based on [https://github.com/vbalnt/tfeat](https://github.com/vbalnt/tfeat).

Additionally, it implements the TripletLoss from Chainer using the autograd tools.

**TODO list:**
- [x]  Replace MNIST dataset by [PhotoTour](http://phototour.cs.washington.edu/patches/default.htm)
- [x] Include testing step and evaluation metrics [FPR95](https://github.com/hanxf/matchnet/blob/master/eval_metrics.py)
- [x] Serialization
- [x] Improve documentation
- [x] Integrate TripletLoss in functional API
- [x] Integrate PhotoTour in `torchvision'


triplet/main.py
```diff
@@ -0,0 +1,177 @@
...
+        print('Generating triplets ...')
+        self.n_triplets = args.n_triplets
+        self.train_triplets = self.generate_triplets(self.train_labels)
+
+    def generate_triplets(self, labels):
+        triplets = []
+        for x in xrange(self.n_triplets):
+            idx = np.random.randint(0, labels.size(0))
+            idx_matches = np.where(labels.numpy() == labels[idx])[0]
+            idx_no_matches = np.where(labels.numpy() != labels[idx])[0]
```
apaszke: You could compute these two arrays for every class outside of the loop. That would likely make it faster.

triplet/main.py
```diff
@@ -0,0 +1,177 @@
...
+    def __init__(self, *arg, **kw):
+        super(TripletMNIST, self).__init__(*arg, **kw)
+
+        print('Generating triplets ...')
+        self.n_triplets = args.n_triplets
+        self.train_triplets = self.generate_triplets(self.train_labels)
+
+    def generate_triplets(self, labels):
+        triplets = []
+        for x in xrange(self.n_triplets):
```
apaszke: xrange works only in Python2

triplet/main.py
```diff
@@ -0,0 +1,184 @@
...
+    def forward(self, anchor, positive, negative):
+        dist = torch.sum(
+            (anchor - positive) ** 2 - (anchor - negative) ** 2,
+            dim=1) + self.margin
+        dist_hinge = torch.clamp(dist, min=0.0)  # maximum between 'dist' and 0.0
+        loss = torch.mean(dist_hinge)
+        return loss
+
+
+def triplet_loss(input1, input2, input3, margin=1.0):
```
apaszke: You can just paste `TripletMarginLoss.forward` code into this function.

apaszke: Then, the module could be removed

triplet/main.py
```diff
@@ -151,7 +145,13 @@ def convert_and_transform(img, transform):
...
 
     def __len__(self):
-        return self.triplets.shape[0]
+        try:
+            if self.train:
+                return self.triplets.shape[0]
+            else:
+                return self.matches.shape[0]
+        except:
+            pass
```
edgarriba: @apaszke I had to hack this since `TripletPhototour` inherits from `Phototour` and for this reason the parent class doesn't have the train attribute which makes crash the program. Any suggestion?

apaszke: No I'm sorry, but I don't think I understand the problem. This `__len__` will never be called by the parent class, what's the error you're getting? I think it should be safe to remove the `try` block

triplet/main.py
```diff
@@ -230,6 +238,31 @@ def train(epoch):
...
+    model.eval()
+    scores, distances = [], []
+    for batch_idx, (data_a, data_p, label) in enumerate(test_loader):
+        if args.cuda:
+            data_a, data_p = data_a.cuda(), data_p.cuda()
+        data_a, data_p, labels = Variable(data_a, volatile=True), \
+                                 Variable(data_p, volatile=True), Variable(label)
+        out_a, out_p = model(data_a), model(data_p)
+        dists = torch.sqrt(torch.sum((out_a - out_p) ** 2, 1))  # euclidean distance
+        distances.extend(dists)
```
edgarriba: @apaszke I don't know why `dists` cannot call `numpy()` which then makes a bit difficult to compute the `ErrorRateAt95Recall`

triplet/main.py
```diff
@@ -230,6 +238,31 @@ def train(epoch):
...
+        scores.extend(label.numpy())
+        if batch_idx % args.log_interval == 0:
+            print('Test Epoch: {} [{}/{} ({:.0f}%)]'.format(
+                epoch, batch_idx * len(data_a), len(test_loader.dataset),
+                100. * batch_idx / len(test_loader)))
+
+        if batch_idx > 70:
+            break
```
edgarriba: @apaszke I had to hack this otherwise I got the follwing error:
```
assert total % total_sizes == 0, "Can't make sizes have exactly %d elements" % total
AssertionError: Can't make sizes have exactly 829440 elements
```

edgarriba: Also during test the gpu is not running

triplet/main.py
edgarriba: That's the error
```
File "/home/eriba/software/pytorch/examples-edgarriba/triplet/main.py", line 148, in __len__
    if self.train:
AttributeError: 'TripletPhotoTour' object has no attribute 'train'
```

I know it's weird, but I traced the error and it looks like that when the parent class is constructed it calls this method instead of it's own

apaszke: Can you share a full stack trace? I need to see where this method was called from.

triplet/main.py
cumttang: The torch Tensor must be transformed into .data, and then it will work.

triplet/main.py
```diff
@@ -230,6 +238,31 @@ def train(epoch):
...
+def test(epoch):
+    model.eval()
+    scores, distances = [], []
+    for batch_idx, (data_a, data_p, label) in enumerate(test_loader):
+        if args.cuda:
+            data_a, data_p = data_a.cuda(), data_p.cuda()
+        data_a, data_p, labels = Variable(data_a, volatile=True), \
+                                 Variable(data_p, volatile=True), Variable(label)
+        out_a, out_p = model(data_a), model(data_p)
+        dists = torch.sqrt(torch.sum((out_a - out_p) ** 2, 1))  # euclidean distance
```
apaszke: You can use `out_a.data.dist(out_p.data)` to compute the euclidean distance.

triplet/main.py
apaszke: @cumttang Good catch. The problem in here, is that models return `Variable`, while you want to work with `Tensors` (that can be obtained with `.data`). Also, are you sure you want to use `.extend` and not `.append`? It will iterate over the tensors, which can be a bit slow.

triplet/main.py
edgarriba: @apaszke I've updated to master version and somehow the error doesn't raise anymore

triplet/main.py
edgarriba: @apaszke any idea how to fix or workaround ?

apaszke: I'll try to reproduce that today. Alternatively can you give me the full error message?

edgarriba:
```
Test Epoch: 1 [99840/100000 (100%)]
  File "/home/eriba/software/pytorch/examples-edgarriba/triplet/main.py", line 268, in <module>
    test(epoch)
  File "/home/eriba/software/pytorch/examples-edgarriba/triplet/main.py", line 249, in test
    out_a, out_p = model(data_a), model(data_p)
  File "/usr/local/lib/python2.7/dist-packages/torch/nn/modules/module.py", line 210, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/eriba/software/pytorch/examples-edgarriba/triplet/main.py", line 168, in forward
    x = x.view(-1, 64*8*8)
  File "/usr/local/lib/python2.7/dist-packages/torch/autograd/variable.py", line 480, in view
    return View(*sizes)(self)
  File "/usr/local/lib/python2.7/dist-packages/torch/autograd/_functions/tensor.py", line 96, in forward
    result = i.view(*self.sizes)
  File "/usr/local/lib/python2.7/dist-packages/torch/tensor.py", line 204, in view
    sizes = _infer_sizes(sizes, self.nelement())
  File "/usr/local/lib/python2.7/dist-packages/torch/tensor.py", line 20, in _infer_sizes
    assert total % total_sizes == 0, "Can't make sizes have exactly %d elements" % total
AssertionError: Can't make sizes have exactly 165888 elements

Process finished with exit code 1
```

edgarriba: it looks like that test batch size must be multiple of the size of the network out (128)

edgarriba: @apaszke Solved the issue in https://github.com/pytorch/examples/pull/47/commits/9d699a0762c7c7c05619dc0ce246a85a9a1373fe#diff-1357f5fd365db925427f5f0692df5b52R161

However, I don't know why before it wasn't working

triplet/main.py
edgarriba: @apaszke here `dist()` won't work since it returns one single value and what I want is the euclidean distance between each of the descriptors in the batch

edgarriba:
@apaszke I think that the code is somehow functional and pretty close to the original implementation with LuaTorch. However, not sure why the loss get stuck at the value of the margin.

/cc @vbalnt Did you find similar issues during the implementation?

triplet/main.py
edgarriba: @cumttang @apaszke it worked. Thanks!

vbalnt:
@edgarriba no such issues with the LuaTorch version. Training was smooth with both margin and ratio criterions.

edgarriba:
@apaszke Are you aware of any difference between optimizers using Lua and Python versions? The same code with Lua is able to achieve a 7% of FPR5 with a LR 0.1, LRD 1e-6 and WD 1e-4. I've tried several LR and the networks always get stuck at ~16% FPR95.

Besides, I've implemented the code in TF which in v0.11 (not in v.12) with MomentumOptimizer M 0.9 and LR 1e-4 I can achieve a 9% FPR95.

apaszke:
Yes, all optimizers were cross-tested with their Lua versions and on toy problem yield the exact same sequence of updates up to 1e-10. It's likely a bug in the implementation somewhere. I'll try to take a look when I'll have a moment

apaszke:
Can you point me to the Lua code?

fmassa:
The only difference I remember between optimizers are different initializations for some parameters. For example [`dampening` in SGD](https://github.com/pytorch/pytorch/issues/6) and the [initialization of the buffers in `RMSProp`](https://github.com/pytorch/pytorch/pull/485)

edgarriba:
@apaszke https://github.com/vbalnt/pnnet/blob/master/train/run.lua
notice that you just need to replace [line 166](https://github.com/vbalnt/pnnet/blob/master/train/run.lua#L116) by nn.MarginRankingCriterion in order to have same approach.

@fmassa hi! :smile: Yeah I noticed about this dampening parameter, in Lua seems that takes momentum value by default. Also in case of defining a learning rate decay, by default this policy it's automatically applied. In Pytorch it's not. As for RMSProp I haven't  tried since I want to replicate original codes where we just use SGD with momentum. Actually, it's a very simple code that should work very easily!

I've tried this tricks but no luck. It looks to me that hyperparameters have to be completely different.

fmassa:
@edgarriba Hi !
I didn't know about the code you sent, nor have I seen the paper, but I have the impression that there are some differences with your code here and the lua code you sent.
For example, I have the impression that [some parts in here](https://github.com/vbalnt/pnnet/blob/master/train/run.lua#L56-L111) is not present in your code, but maybe that's because you have optimized some operations out in your `TripletLoss`?
The [Min](https://github.com/vbalnt/pnnet/blob/master/train/run.lua#L100) over two distances seems to be absent (except if it's already mixed somehow with the MarginRankingCriterion an present in [here](https://github.com/pytorch/examples/pull/47/files#diff-1357f5fd365db925427f5f0692df5b52R219))?

edgarriba:
@fmassa The code is slightly different since with Pytorch you can easily share the weights of the net (or at least is what I assumed) whereas in the Lua version you had to deal with `ConcatTable` in order to define such architectures.

As you said some of the parts are optimized inside the loss function like the two `PairDistance` functions. In this case basically I took the TripletLoss from Chainer and I adapted to Pytorch routines. Ideally should be the same as the one that I implemented with TensorFlow¹ which btw using it I achieve almost the same results reported to the paper.

What you mention about the min is an extra contribution that we introduced in the paper² in order to improve the loss function.

¹ https://github.com/vbalnt/tfeat/blob/master/tensorflow/train/tfeat/loss.py#L13-L35
² http://www.iis.ee.ic.ac.uk/%7Evbalnt/shallow_descr/TFeat_paper.pd

fmassa:
@edgarriba I didn't spend much time looking into it, but I thought that in the lua code there were 3 pairwise distances, while in the Pytorch implementation you only compute 2 pairwise distances. But maybe I'm missing more things

edgarriba:
@fmassa Yes 3 pairwise distances are for the contribution in the paper, but for the triplet base case just two are needed. I'm trying the base case.

edgarriba:
@apaszke could you revise again this?
Before merging I'd to in integrate PairwiseDistance and both Triplet losses into the main repo.
I guess here? https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/distance.py

Also for the Phototour dataset I'll send a PR to
https://github.com/pytorch/vision/tree/master/torchvision/datasets

However, by the moment I assume that the user has opencv installed since there's a big gap in terms of speed compared to PIL. How do you want to handle this?

triplet/main.py
```diff
@@ -0,0 +1,487 @@
...
+
+    def forward(self, x1, x2):
+        assert x1.size() == x2.size()
+        eps = 1e-4 / x1.size(1)
+        diff = torch.abs(x1 - x2)
+        out = torch.pow(diff, self.norm).sum(dim=1)
+        return torch.pow(out + eps, 1. / self.norm)
+
+
+class TripletMarginLoss(Function):
```
apaszke: You should subclass a `Function` only if you want to extend autograd, which is not what you're doing in here (you haven't defined `backward`, so it's not a valid one). If you want to compose some autograd ops either use a regular Python function declared with `def`, or subclass from `object` or `nn.Module`

Same goes for `TripletMarginLossSwap` below.

triplet/main.py
```diff
@@ -0,0 +1,487 @@
...
+args.cuda = not args.no_cuda and torch.cuda.is_available()
+torch.manual_seed(args.seed)
+np.random.seed(args.seed)
+cv2.setRNGSeed(args.seed)
+
+if not os.path.exists(args.log_dir):
+    os.makedirs(args.log_dir)
+
+if args.cuda:
+    cudnn.enabled = True
```
apaszke: No need for that.

triplet/main.py
```diff
@@ -0,0 +1,487 @@
...
+    cudnn.benchmark = True
+    torch.cuda.manual_seed(args.seed)
```
apaszke: `manual_seed_all` will already do that for you. Just remove this line

triplet/main.py
```diff
@@ -0,0 +1,487 @@
...
+            triplets.append([indices[c1][n1], indices[c1][n2], indices[c2][n3]])
+        return torch.LongTensor(np.array(triplets))
+
+    def __getitem__(self, index):
+        def transform(img):
+            """Convert image into numpy array and apply transformation
+               Doing this so that it is consistent with all other datasets
+               to return a PIL Image.
+            """
+            if transform is not None:
```
apaszke: How would it ever be `None`? It's going to be a reference to that function.

triplet/main.py
```diff
@@ -0,0 +1,487 @@
...
+        )
+
+    def forward(self, x):
+        x = self.features(x)
+        x = x.view(x.size(0), -1)
+        x = self.classifier(x)
+        return x
+
+
+def weights_init(m):
```
apaszke: This is going to get merged into core soon.

triplet/main.py
apaszke: Please paste that code into this function. Moving it out to these separate classes doesn't add anything.

triplet/main.py
edgarriba: @apaszke see below, is it correct like this?

edgarriba:
ping @apaszke

bartolsthoorn:
I think it would be cool to have the `PhotoTour` as a triplet `Sampler` instance. I mean something similar to the ones here: [sampler.py](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py) You can pass those to a DataLoader. It could for example be called `TripletSampler`.

triplet/eval_metrics.py
```diff
@@ -0,0 +1,28 @@
...
+https://github.com/hanxf/matchnet/blob/master/eval_metrics.py
+"""
+import operator
+
+
+def ErrorRateAt95Recall(labels, scores):
+    recall_point = 0.95
+    # Sort label-score tuples by the score in descending order.
+    sorted_scores = zip(labels, scores)
+    sorted_scores.sort(key=operator.itemgetter(1), reverse=False)
```
ClementPinard: This line does not work for python3, as zip now returns an iterator.
see here : https://stackoverflow.com/questions/43698531/sorting-a-zipped-object-in-python-3

So i suppose it's better to either have 
```python
sorted_scores = list(zip(labels, scores))
sorted_scores.sort(key=operator.itemgetter(1), reverse=False)
```
or (the best IMO)
```python
sorted_scores = zip(labels, scores)
sorted_scores = sorted(sorted_scores, key=operator.itemgetter(1), reverse=False)
```

triplet/main.py
```diff
@@ -0,0 +1,391 @@
...
+
+    optimizer = create_optimizer(model, args.lr)
+
+    # optionally resume from a checkpoint
+    if args.resume:
+        if os.path.isfile(args.resume):
+            print('=> loading checkpoint {}'.format(args.resume))
+            checkpoint = torch.load(args.resume)
+            args.start_epoch = checkpoint['epoch']
+            checkpoint = torch.load(args.resume)
```
gzuidhof: You load the checkpoint twice, I think once will work too.

posix4e:
What's going on here?

msaroufim:
Closing since WIP tag since 2017

PR #60, opened 2017-01-27
Fast Neural Style example

clcarwin:
Port from https://github.com/yusuketomoto/chainer-fast-neuralstyle


soumith:
cc: @jcjohnson

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+        # h  = F.relu(self.conv5_2(h))
+        # h  = F.relu(self.conv5_3(h))
+        # h  = F.max_pool2d(h, kernel_size=2, stride=2)
+
+        return [f1,f2,f3,f4] #[relu1_2,relu2_2,relu3_3,relu4_3]
+
+class ResidualBlock(nn.Module):
+    def __init__(self,num):
+        super(ResidualBlock, self).__init__()
+        self.c1=nn.Conv2d(num, num, kernel_size=3, stride=1, padding=1)
```
jcjohnson: I had bad border artifacts when using zero-padding, so I used reflection padding at the very beginning of the network to ensure that the input and output have the same size.

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+        h = F.relu(self.b2(self.c2(h)))
+        h = F.relu(self.b3(self.c3(h)))
+        h = self.r1(h)
+        h = self.r2(h)
+        h = self.r3(h)
+        h = self.r4(h)
+        h = self.r5(h)
+        h = F.relu(self.b4(self.d1(h)))
+        h = F.relu(self.b5(self.d2(h)))
+        y = self.d3(h)
```
jcjohnson: In the paper I used a tanh after the last convolution to ensure the output from the generator network is bounded.

Since the paper came out I've done some experiments without the tanh and found that it isn't really necessary.

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+class VGG(nn.Module):
```
jcjohnson: Can you use the VGG definition from `torchvision.models`?

fast_neural_style/generate.py
```diff
@@ -0,0 +1,46 @@
...
+import numpy as np
+import os
+import cv2
+
+def load_image(path):
+    img = cv2.imread(path).astype('float32')
+    img = img.transpose(2, 0, 1)
+    return img
+
+parser = argparse.ArgumentParser(description='PyTorch MNIST Example')
```
jcjohnson: Nit: Change description

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+    vgg.cuda()
+
+
+style = load_image(args.style_image,image_size) - 120
+style = style.reshape((1,)+style.shape)
+style = torch.from_numpy(style)
+if(args.cuda):
+    style = style.cuda()
+feature_s = vgg(Variable(style, volatile=True))
+gram_s = [gram_matrix(y) for y in feature_s]
```
jcjohnson: You need to be more careful with image preprocessing. The VGG networks were originally trained in Caffe, so before you feed images to VGG you need to do the following:

(1) Make sure the image is BGR instead of RGB
(2) Make sure the pixels are in the range [0, 255]; most of Torch assumes [0, 1]
(3) Subtract the mean ImageNet pixel: BGR values is [103.939, 116.779, 123.68]

I found it most convenient to let the generator produce images with VGG preprocessing so its output could be fed directly to the loss network; however this means that you need to do explicit deprocessing of the generator output when you actually want to save the output.

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+
+loss_fn = torch.nn.MSELoss()
+model.train()
+for epoch in range(n_epoch):
+    print('epoch', epoch)
+    for i in range(n_iter):
+        optimizer.zero_grad()
+
+        img = load_image(imagepaths[i], image_size)
```
jcjohnson: It seems like you're training with a batch size of 1? This might make convergence harder. I used a batch size of 4 in the paper.

Also, to better showcase PyTorch features, you should consider using a PyTorch DataLoader to load data; you can easily load images off disk using a `torchvision.models.ImageFolder` dataset. Aside from being more idiomatic PyTorch, this will also give you multithreaded data loading for free!

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+model = models.FastStyleNet()
+vgg = models.VGGFeature()
+vgg.load_state_dict(torch.load('vgg16feature.pth'))
+optimizer = optim.Adam(model.parameters(), lr=args.lr)
+if args.cuda:
+    model.cuda()
+    vgg.cuda()
+
+
+style = load_image(args.style_image,image_size) - 120
```
jcjohnson: You should have a separate hyperparameter controlling the size of the style image. This ends up being a very important parameter for controlling the look of the stylized images.

fast_neural_style/net.py
clcarwin: We can change ```models.vgg16feature()``` to```torchvision.models.vgg16()``` to convert a torch vgg16 model to a pytorch version. In this example, we have to use ```models.vgg16feature()``` for exporting  4 features ```relu1_2,relu2_2,relu3_3,relu4_3```.
```python
def download_convert_vgg16_model():
       ...
        vgg = models.vgg16feature()
        for src,dst in zip(vgglua.parameters()[0],vgg.parameters()):
            dst[:] = src[:]
        torch.save(vgg.state_dict(), 'vgg16feature.pth')
```

fast_neural_style/net.py
clcarwin: That's a good idea. In my practice, I add a 50x50 border and crop it in result.

fast_neural_style/net.py
clcarwin: Yes, I also can not find more benefit of the tanh.

fast_neural_style/train.py
clcarwin: OK. A ```--style_size``` parameter will be added.

fast_neural_style/train.py
clcarwin: Thanks for your advice.

1. I have used ```cv2.imread``` to load image and opencv's image defaults to BGR.
2. The range of opencv's image defaults to [0,255].
3. I have used [120,120,120] to replace the mean [103.939, 116.779, 123.68]. The preprocessing can be reduced to just one line ```x = x - 120```, the deprocessing can be reduced to ```x = x + 120```. It may be more easy for some starters to read the whole code. After this simplification, there is no any visible difference in the output image.

fast_neural_style/train.py
clcarwin: Yes, a batch size of 1 is used. It is for a shorter and simpler code. I can add a batch size parameter later.

In this case, we can not use ```torchvision.datasets.ImageFolder```. Because the **target/label** is not static, it is calculated by vgg model dynamically. In addition, ```ImageFolder``` use PIL, we will add many code to deal with RGB/BGR and [0,255]/[0,1].

clcarwin:
These problems will be fixed in next few days. If there's anything else, please feel free to tell me.

1. Add two parameter for train.py ```--style_size``` ```--batch_size```.
2. Fix argparse's description in generate.py.
3. Add some pretrained models.

fast_neural_style/generate.py
```diff
@@ -0,0 +1,46 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from torchvision import datasets, transforms
+from torch.autograd import Variable
+
+import net as models
+import numpy as np
+import os
+import cv2
```
apaszke: Is really needed? I think all functionality you need is already implemented as `torchvision.transforms`.

fast_neural_style/generate.py
```diff
@@ -0,0 +1,46 @@
...
+    img = cv2.imread(path).astype('float32')
+    img = img.transpose(2, 0, 1)
+    return img
+
+parser = argparse.ArgumentParser(description='PyTorch MNIST Example')
+parser.add_argument('--no-cuda', action='store_true', default=False,
+                    help='enables CUDA training')
+parser.add_argument('--model', default='model.pth', type=str)
+parser.add_argument('--input', type=str, required=True)
+parser.add_argument('--output', default='output.jpg', type=str)
```
apaszke: Can you please add help for arguments?

fast_neural_style/generate.py
```diff
@@ -0,0 +1,46 @@
...
+
+model = models.FastStyleNet()
+model.load_state_dict(torch.load(args.model))
+
+img = load_image(args.input)
+img = img.reshape((1,)+img.shape)
+img = torch.from_numpy(img)
+img = Variable(img, volatile=True)
+
+if(args.cuda):
```
apaszke: nit: parenthesis

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return [f1,f2,f3,f4] #[relu1_2,relu2_2,relu3_3,relu4_3]
+
+class ResidualBlock(nn.Module):
+    def __init__(self,num):
+        super(ResidualBlock, self).__init__()
+        self.c1=nn.Conv2d(num, num, kernel_size=3, stride=1, padding=1)
+        self.c2=nn.Conv2d(num, num, kernel_size=3, stride=1, padding=1)
+        self.b1=nn.BatchNorm2d(num)
+        self.b2=nn.BatchNorm2d(num)
```
apaszke: nit: add spaces before and after `=`.

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+        self.r4=ResidualBlock(128)
+        self.r5=ResidualBlock(128)
+        self.d1=nn.ConvTranspose2d(128, 64, kernel_size=4, stride=2, padding=1)
+        self.d2=nn.ConvTranspose2d(64, 32, kernel_size=4, stride=2, padding=1)
+        self.d3=nn.Conv2d(32, 3, kernel_size=9, stride=1, padding=4)
+        self.b1=nn.BatchNorm2d(32)
+        self.b2=nn.BatchNorm2d(64)
+        self.b3=nn.BatchNorm2d(128)
+        self.b4=nn.BatchNorm2d(64)
+        self.b5=nn.BatchNorm2d(32)
```
apaszke: As above

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+        h = self.r4(h)
+        h = self.r5(h)
+        h = F.relu(self.b4(self.d1(h)))
+        h = F.relu(self.b5(self.d2(h)))
+        y = self.d3(h)
+        return y
+
+
+def vgg16():
+    return VGG()
```
apaszke: Is this net even used anywhere?

fast_neural_style/net.py
```diff
@@ -0,0 +1,165 @@
...
+        h  = F.relu(self.conv4_3(h))
+        f4 = h
+        # h  = F.max_pool2d(h, kernel_size=2, stride=2)
+
+        # h  = F.relu(self.conv5_1(h))
+        # h  = F.relu(self.conv5_2(h))
+        # h  = F.relu(self.conv5_3(h))
+        # h  = F.max_pool2d(h, kernel_size=2, stride=2)
+
+        return [f1,f2,f3,f4] #[relu1_2,relu2_2,relu3_3,relu4_3]
```
apaszke: I think it's better to reconstruct this net, by iterating over the modules and slicing them at each pooling layer. Then you insert all slices as sequentials and write custom forward that gathers the states after each one. I can give you the code if you want.

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+    img = cv2.imread(path).astype('float32')
+    img = cv2.resize(img,(size,size))
+    img = img.transpose(2, 0, 1)
+    return img
+
+def gram_matrix(y):
+    b, ch, h, w = y.size()
+    features = y.view(b, ch, w*h)
+    features_t = features.transpose(1,2)
+    gram = features.bmm(features_t)/(ch*h*w)
```
apaszke: nit: spaces

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+optimizer = optim.Adam(model.parameters(), lr=args.lr)
+if args.cuda:
+    model.cuda()
+    vgg.cuda()
+
+
+style = load_image(args.style_image,image_size) - 120
+style = style.reshape((1,)+style.shape)
+style = torch.from_numpy(style)
+if(args.cuda):
```
apaszke: nit: parenthesis

fast_neural_style/train.py
apaszke: Yeah, but I have no doubt that the existing users will be asking why is it using incorrect preprocessing. I'd fix that, or at least leave a comment.

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+        xc = Variable(x.clone(), volatile=True)
+        x = Variable(x)
+        y = model(x)
+
+        xc -= 120
+        y -= 120
+        feature = vgg(xc)
+        feature_hat = vgg(y)
+
+        L_feat = lambda_f * loss_fn(feature_hat[2], Variable(feature[2].data,requires_grad=False)) #relu3_3
```
apaszke: nit: spaces

fast_neural_style/train.py
```diff
@@ -0,0 +1,127 @@
...
+        L_sytle = 0.0
```
apaszke: typo

fast_neural_style/train.py
apaszke: You can give a custom loader function to the `DataLoader`. I can't see why it wouldn't work here.

fast_neural_style/net.py
clcarwin: If you can provide the sample code, it would be very great. Thanks.

clcarwin:
@apaszke @jcjohnson 
Hi, all requested changes have been changed. Please review it again. Thanks.

szagoruyko:
@clcarwin nice work! I added the models to the functional zoo here https://github.com/szagoruyko/functional-zoo/blob/master/fast-neural-style.ipynb

transedward:
@clcarwin something irrelevant with this PR.
I saw you manage to remove the dependency with `cv2`.
I am especially interested in how you replace `cv2.resize` functionality with something in `torchvision`.

Sorry I'm not really familiar with vision but also want to remove dependency on `cv2`

darkstar112358:
@apaszke 

Hi, I also implemented a Fast Neural Style in pytorch. Most of the implementation is similar to the implementation present in this pr. I made a few changes like replacing batch-normalization with instance-normalization which gives better results.

repo: https://github.com/darkstar112358/fast-neural-style

I can work on adding fast-neural-style to pytorch/examples if you need it.

soumith:
this has been (unfortunately) succeeded by https://github.com/pytorch/examples/pull/129 which is now merged into master. Thanks a lot for the PR.

PR #65, opened 2017-01-30
Fix adjust learning rate

Junonia:
Fix adjust learning rate, the original code does not work with the latest pytorch


soumith:
thanks!

PR #66, opened 2017-02-07
Updating with torchtext changes and snapshot loading

bmccann:
Added loading from snapshot
Updated lowercasing through torchtext
Addressed PyTorch issue https://github.com/pytorch/pytorch/issues/689 with  torch.cuda.set_device(args.gpu)


soumith:
thanks!

PR #69, opened 2017-02-16
Prevent 2 forward passes with detach

Kaixhin:
Similarly to the Torch example, this reuses the fake image that was created for the discriminator (and hence the forward pass of the generator).

Checked that the expected backwards passes are occurring using hooks, and have also run the example to check that it's working fine empirically.


apaszke:
Thanks!

PR #72, opened 2017-02-20
Mnist example using specified GPU

wkentaro:
Currently it does not only use specified gpu, but also use gpu id 0 (as shown below).
Is this a expected behavior, or am I missing something?
I also tried passing `args.gpu` to all `.cuda()` function like `model.cuda(args.gpu)`, but it does not change the behavior compared to that using `torch.cuda.set_device(args.gpu)`.

**gpu id == 1**

```
% python main.py --gpu 1

% nvidia-smi
Tue Feb 21 04:25:45 2017
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 367.57                 Driver Version: 367.57                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  TITAN X (Pascal)    Off  | 0000:05:00.0      On |                  N/A |
| 23%   36C    P8    12W / 250W |    459MiB / 12188MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  TITAN X (Pascal)    Off  | 0000:06:00.0     Off |                  N/A |
| 23%   39C    P8    17W / 250W |      3MiB / 12189MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   2  TITAN X (Pascal)    Off  | 0000:09:00.0     Off |                  N/A |
| 23%   45C    P2    58W / 250W |    380MiB / 12189MiB |     19%      Default |
+-------------------------------+----------------------+----------------------+
|   3  TITAN X (Pascal)    Off  | 0000:0A:00.0     Off |                  N/A |
| 23%   31C    P8    16W / 250W |    288MiB / 12189MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|    0      1255    G   /usr/lib/xorg/Xorg                             230MiB |
|    0      2874    G   compiz                                         216MiB |
|    0      8586    G   ...nDeveloperModeWarning/Default/Html5ByDefa     8MiB |
|    2     26563    C   python                                         377MiB |
|    3     26563    C   python                                         285MiB |
+-----------------------------------------------------------------------------+
```

[...]

PR #73, opened 2017-02-21
Add instruction to use specified GPU id

wkentaro:
Related to #72
Is this information available in docs?


soumith:
this information is available in the CUDA documentation.

wkentaro:
Thanks!

PR #76, opened 2017-02-23
fixed divide by zero error

atgambardella:
If the agent never experienced reward, rewards turns into nan (0/0). We should add machine epsilon to the denominator here to avoid that.


soumith:
thanks!

PR #78, opened 2017-02-23
fix encStates on translateBatch after src encoding

geovedi:
got size mismatch during translation when using bidirectional model. this should fix the problem.


soumith:
i'm sorry, i merged in https://github.com/pytorch/examples/pull/80 by @bmccann just now. Can you rebase / is the fix still relevant?

geovedi:
no, no longer relevant as #80 has the fix.

PR #80, opened 2017-02-23
Multi-GPU OpenNMT

bmccann:
Changes necessary to use OpenNMT with nn.DataParallel.
Reference https://github.com/adamlerer/examples/pull/1

-  While it does train correctly, additional GPUs appear to slow down training. Perhaps this is due to issues similar to those towards the end of https://discuss.pytorch.org/t/does-it-support-multi-gpu-card-on-a-single-node/75. If anyone has more familiarity with using nn.DataParallel with RNNs, I'd love to chat about how to remedy this (or at least figure out a more precise diagnosis).

- There's likely a cleaner way to refactor the code to avoid so many transposes. I initially did not want to change the logic that was used in the original batching code to ensure correctness, but this is something I can look into further if the transposes are an issue.

### DIFF
--- a/OpenNMT/onmt/Models.py
+++ b/OpenNMT/onmt/Models.py
@@ -110,18 +110,18 @@ class Decoder(nn.Module):
         # self.input_feed=False
         outputs = []
         output = init_output
-        for emb_t in emb.chunk(emb.size(0)):
+        for i, emb_t in enumerate(emb.chunk(emb.size(0), dim=0)):
             emb_t = emb_t.squeeze(0)
             if self.input_feed:
                 emb_t = torch.cat([emb_t, output], 1)
 
-            output, hidden = self.rnn(emb_t, hidden)
+            output, h = self.rnn(emb_t, hidden)
             output, attn = self.attn(output, context.t())
             output = self.dropout(output)
             outputs += [output]
 
         outputs = torch.stack(outputs)
-        return outputs.transpose(0, 1), hidden, attn
+        return outputs.transpose(0, 1), h, attn
 
 
 class NMTModel(nn.Module):
--- a/OpenNMT/onmt/Models.py
+++ b/OpenNMT/onmt/Models.py
@@ -29,10 +29,9 @@ class Encoder(nn.Module):
             self.word_lut.weight.copy_(pretrained)
 
     def forward(self, input, hidden=None):
-        emb = self.word_lut(input)
-
+        batch_size = input.size(0) # batch first for multi-gpu compatibility
+        emb = self.word_lut(input).transpose(0, 1)
         if hidden is None:
-            batch_size = emb.size(1)
             h_size = (self.layers * self.num_directions, batch_size, self.hidden_size)
             h_0 = Variable(emb.data.new(*h_size).zero_(), requires_grad=False)
             c_0 = Variable(emb.data.new(*h_size).zero_(), requires_grad=False)
@@ -46,21 +45,21 @@ class StackedLSTM(nn.Module):
     def __init__(self, num_layers, input_size, rnn_size, dropout):
         super(StackedLSTM, self).__init__()
         self.dropout = nn.Dropout(dropout)
+        self.num_layers = num_layers
 
-        self.layers = []
         for i in range(num_layers):
             layer = nn.LSTMCell(input_size, rnn_size)
             self.add_module('layer_%d' % i, layer)
-            self.layers += [layer]
             input_size = rnn_size
 
     def forward(self, input, hidden):
         h_0, c_0 = hidden
         h_1, c_1 = [], []
-        for i, layer in enumerate(self.layers):
+        for i in range(self.num_layers):
+            layer = getattr(self, 'layer_%d' % i)
             h_1_i, c_1_i = layer(input, (h_0[i], c_0[i]))
             input = h_1_i
-            if i != len(self.layers):
+            if i != self.num_layers:
                 input = self.dropout(input)
             h_1 += [h_1_i]
             c_1 += [c_1_i]
@@ -99,9 +98,9 @@ class Decoder(nn.Module):
 
 
     def forward(self, input, hidden, context, init_output):
-        emb = self.word_lut(input)
+        emb = self.word_lut(input).transpose(0, 1)
 
-        batch_size = input.size(1)
+        batch_size = input.size(0)
 
         h_size = (batch_size, self.hidden_size)
         output = Variable(emb.data.new(*h_size).zero_(), requires_grad=False)
@@ -122,7 +121,7 @@ class Decoder(nn.Module):
             outputs += [output]
 
         outputs = torch.stack(outputs)
-        return outputs, hidden, attn
+        return outputs.transpose(0, 1), hidden, attn
 
 
 class NMTModel(nn.Module):
@@ -154,7 +153,7 @@ class NMTModel(nn.Module):
 
     def forward(self, input):
         src = input[0]
-        tgt = input[1][:-1]  # exclude last target from inputs
+        tgt = input[1][:, :-1]  # exclude last target from inputs
         enc_hidden, context = self.encoder(src)
         init_output = self.make_init_decoder_output(context)
 
--- a/OpenNMT/train.py
+++ b/OpenNMT/train.py
@@ -25,9 +25,9 @@ parser.add_argument('-train_from',
 
 parser.add_argument('-layers', type=int, default=2,
                     help='Number of layers in the LSTM encoder/decoder')
-parser.add_argument('-rnn_size', type=int, default=512,
+parser.add_argument('-rnn_size', type=int, default=500,
                     help='Size of LSTM hidden states')
-parser.add_argument('-word_vec_size', type=int, default=300,
+parser.add_argument('-word_vec_size', type=int, default=500,
                     help='Word embedding sizes')
 parser.add_argument('-input_feed', type=int, default=1,
                     help="""Feed the context vector at each time step as
@@ -43,13 +43,13 @@ parser.add_argument('-brnn_merge', default='concat',
 
 ## Optimization options
 
-parser.add_argument('-batch_size', type=int, default=256,
+parser.add_argument('-batch_size', type=int, default=64,
                     help='Maximum batch size')
 parser.add_argument('-max_generator_batches', type=int, default=32,
                     help="""Maximum batches of words in a sequence to run
                     the generator on in parallel. Higher is faster, but uses
                     more memory.""")
-parser.add_argument('-epochs', type=int, default=50,
+parser.add_argument('-epochs', type=int, default=13,
                     help='Number of training epochs')
 parser.add_argument('-start_epoch', type=int, default=1,
                     help='The epoch from which to start')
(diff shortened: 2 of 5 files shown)


soumith:
cc: @ngimel who would be interested about (1)

ngimel:
@bmccann, what's the command lines to compare multi-GPU and single-GPU performance?

bmccann:
@ngimel 

I'm using` python train.py -data data/demo-train.pt -gpus 0` for single gpu
and` python train.py -data data/demo-train.pt -gpus 0 1` for multi-gpu

OpenNMT/onmt/Models.py
```diff
@@ -46,21 +45,21 @@ class StackedLSTM(nn.Module):
...
             self.add_module('layer_%d' % i, layer)
-            self.layers += [layer]
             input_size = rnn_size
 
     def forward(self, input, hidden):
         h_0, c_0 = hidden
         h_1, c_1 = [], []
-        for i, layer in enumerate(self.layers):
+        for i in range(self.num_layers):
+            layer = getattr(self, 'layer_%d' % i)
```
adamlerer: I think you can use `nn.ModuleList` for this now.

OpenNMT/onmt/Models.py
```diff
@@ -111,18 +110,18 @@ def forward(self, input, hidden, context, init_output):
         # self.input_feed=False
         outputs = []
         output = init_output
-        for emb_t in emb.chunk(emb.size(0)):
+        for i, emb_t in enumerate(emb.chunk(emb.size(0), dim=0)):
```
adamlerer: `emb.split(1)` is a better choice now.

OpenNMT/onmt/Models.py
```diff
@@ -99,9 +98,9 @@ def __init__(self, opt, dicts):
 
 
     def forward(self, input, hidden, context, init_output):
-        emb = self.word_lut(input)
+        emb = self.word_lut(input).transpose(0, 1)
```
adamlerer: If you're going to be operating as (batch, seq_len, dim) everywhere, maybe it's better to just have the data in this format end-to-end? I.e. starting with the data loader. I'm not sure if this is better, it requires more far-reaching changes.

OpenNMT/train.py
```diff
@@ -95,11 +96,15 @@
 #                     help="Seed for random initialization")
 
 opt = parser.parse_args()
+opt.cuda = len(opt.gpus)
```
adamlerer: Can you remove this? You can use `opt.gpus` directly for bool comparisons (since [] evaluates to false), and use `len(opt.gpus)` directly elsewhere.

OpenNMT/train.py
```diff
@@ -151,10 +156,14 @@ def eval(model, criterion, data):
 def trainModel(model, trainData, validData, dataset, optim):
     print(model)
     model.train()
+    if optim.last_ppl is None:
```
adamlerer: Why is this necessary? We initialize the parameters below, no?

OpenNMT/onmt/Models.py
bmccann: Yes, that didn't exist when I wrote this. I'll update.

OpenNMT/onmt/Models.py
```diff
@@ -29,10 +29,9 @@ def __init__(self, opt, dicts):
             self.word_lut.weight.copy_(pretrained)
 
     def forward(self, input, hidden=None):
-        emb = self.word_lut(input)
-
+        batch_size = input.size(0) # batch first for multi-gpu compatibility
+        emb = self.word_lut(input).transpose(0, 1)
```
bmccann: I'll run that comparison.

OpenNMT/onmt/Models.py
bmccann: I agree that an end-to-end change would be better. I was hoping for your feedback on this before messing with your setup too much, but now that it is merged, I'll make that a priority with the speed comparisons.

OpenNMT/train.py
bmccann: Will do

OpenNMT/train.py
bmccann: That should not be there. Must have gotten mistakenly carried over in a rebase.

OpenNMT/translate.py
```diff
@@ -48,6 +49,8 @@ def reportScore(name, scoreTotal, wordsTotal):
 
 def main():
     opt = parser.parse_args()
+    opt.cuda = opt.gpu > -1
+    torch.cuda.set_device(opt.gpu)
```
bmccann: Bug

bmccann:
@adamlerer 

Multi-GPU: 

```
Epoch  1,    50/  156 batches; perplexity: 99443.77; 2330 tokens/s;     31 s elapsed
Epoch  1,   100/  156 batches; perplexity: 8743.52; 2506 tokens/s;     59 s elapsed
Epoch  1,   150/  156 batches; perplexity: 4907.59; 2610 tokens/s;     88 s elapsed
Train perplexity: 15371.7
Validation perplexity: 5809.43
```

Single-GPU:

```
Epoch  1,    50/  156 batches; perplexity: 169375.25; 2719 tokens/s;     27 s elapsed
Epoch  1,   100/  156 batches; perplexity: 28906.83; 2715 tokens/s;     52 s elapsed
Epoch  1,   150/  156 batches; perplexity: 6529.42; 2678 tokens/s;     81 s elapsed
Train perplexity: 29324.4
Validation perplexity: 4475.37
```

OpenNMT/onmt/Models.py
bmccann: Backing up to before this PR was merged, looks like it ran about 300 tokens/s faster than after the merge (with all the transposes). I'll work on making everything batch first, and see if that makes up the difference.

bmccann: Before this PR: Epoch  1,    50/  156 batches; perplexity: 169995.82; 2826 tokens/s

```Epoch  1,   100/  156 batches; perplexity: 29357.66; 3026 tokens/s
Epoch  1,   150/  156 batches; perplexity: 7109.44; 3126 tokens/s
Train perplexity: 27737.7
Validation perplexity: 2487.7

Epoch  2,    50/  156 batches; perplexity: 3122.17; 2974 tokens/s
Epoch  2,   100/  156 batches; perplexity: 2255.79; 3118 tokens/s
Epoch  2,   150/  156 batches; perplexity: 1812.20; 2968 tokens/s
Train perplexity: 2311.4
Validation perplexity: 1759
```

and after this PR:

```
Epoch  1,    50/  156 batches; perplexity: 1109740.55; 3038 tokens/s;     26 s elapsed
Epoch  1,   100/  156 batches; perplexity: 14303.51; 3002 tokens/s;     49 s elapsed
Epoch  1,   150/  156 batches; perplexity: 4059.63; 2937 tokens/s;     72 s elapsed
Train perplexity: 43706.5
Validation perplexity: 3200.05

Epoch  2,    50/  156 batches; perplexity: 2723.29; 2962 tokens/s;    103 s elapsed
Epoch  2,   100/  156 batches; perplexity: 1981.58; 3061 tokens/s;    128 s elapsed
Epoch  2,   150/  156 batches; perplexity: 1816.14; 2977 tokens/s;    153 s elapsed
Train perplexity: 2111.29
Validation perplexity: 2029.71
```

OpenNMT/onmt/Models.py
bmccann: The speed difference with the transposes seems pretty negligible on the demo data, but I still think it would be nice to avoid them where possible, so I'll look at switching it over to (batch, seq_len, dim) everywhere.

OpenNMT/onmt/Models.py
adamlerer: Looks like the transposes haven't decreased performance (although in a later comment you said it's 300 words/s slower, so I'm not clear). That's good.

bmccann: Yeah, my bad about the previous comment. I wasn't running exactly the same thing. What I've posted above, I got from rolling back to c84f7df24bc8df556c82908a5fbd186fc38f303e and comparing with master.

adamlerer:
Interesting. Well, if multi-GPU is slower than single-GPU, there's no point in merging this PR until we've figured it out. Have you tried running with the Lua Torch version to see if you get the same behavior?

Some possibilities:
- RNNs are pretty parameter-heavy, so for small batch size maybe you spend all your time communicating parameters
- you're bottlenecked by Python executing the kernels on multiple GPUs since they're so small. I think DataParallel uses multiple threads, but the python parts hold on to the GIL.

cc @colesbury

bmccann:
Happy to leave this as WIP until we make multi-GPU faster. I think we'll still want to take some of the smaller changes from this branch that fix bugs that are in master without it though. https://github.com/bmccann/examples/tree/onmt_small_fixes will make all the changes we've talked about except batch-first end-to-end. If we decide to roll back the merge, I'll move the other bug fixes in this branch to that one so they can merge sooner.

ngimel:
DataParallel at this point does not use multiple threads for backwards, and the code is very CPU bound (as @adamlerer pointed out, kernels are very small). Even threaded forward part remains CPU-bound.  Parameter exchange takes some time, but the biggest problem I see is small kernels. May be running with bigger rnn_size/word_vec_size will have better scaling.

bmccann:
I see. Now that you mention larger batches, etc. I've rerun with a batch size of 512 (everything else left as default), and I'm seeing the multi-gpu code get a bit of a speedup, so for now this may only be useful for larger model/batch combinations.

Single GPU:

```
Epoch  1,    10/   19 batches; perplexity: 563542.92; 3993 tokens/s;     33 s elapsed
Train perplexity: 287798
Validation perplexity: 2.91026e+08

Epoch  2,    10/   19 batches; perplexity: 874205.98; 3957 tokens/s;     91 s elapsed
Train perplexity: 504300
Validation perplexity: 57021

Epoch  3,    10/   19 batches; perplexity: 78948.53; 3867 tokens/s;    149 s elapsed
Train perplexity: 58810.5
Validation perplexity: 13508.7

Epoch  4,    10/   19 batches; perplexity: 79398.57; 3973 tokens/s;    208 s elapsed
Train perplexity: 58670.9
Validation perplexity: 12442.1

Epoch  5,    10/   19 batches; perplexity: 16059.98; 4030 tokens/s;    265 s elapsed
Train perplexity: 10363
Validation perplexity: 15048.1
Decaying learning rate to 0.5
```

Two GPUs:

```
Epoch  1,    10/   19 batches; perplexity: 921818.14; 4610 tokens/s;     29 s elapsed
Train perplexity: 692491
Validation perplexity: 50474.8

Epoch  2,    10/   19 batches; perplexity: 741175.54; 5485 tokens/s;     71 s elapsed
Train perplexity: 217865
Validation perplexity: 42705.7

Epoch  3,    10/   19 batches; perplexity: 49493.03; 5504 tokens/s;    115 s elapsed
Train perplexity: 32258.1
Validation perplexity: 68482.7
Decaying learning rate to 0.5

Epoch  4,    10/   19 batches; perplexity: 6975.58; 5166 tokens/s;    157 s elapsed
Train perplexity: 5871.81
Validation perplexity: 6592.61
Decaying learning rate to 0.25

Epoch  5,    10/   19 batches; perplexity: 3146.31; 5716 tokens/s;    200 s elapsed
Train perplexity: 3232.02
Validation perplexity: 2681.03
Decaying learning rate to 0.125
```

ngimel:
Also, replacing LSTMCell with a single-step single-layer LSTM (that would get routed to cudnn) might help, especially for small hidden sizes. On the attached profile the small kernels after 2 gemms are pointwise operations that would get fused by cudnn (and reduce load on the CPU). That would help single GPU performance as well. Some massaging of inputs and outputs will be required (LSTMCell expects 2D input, LSTM 3D etc). 
selection_009

apaszke:
I think that an additional improvement would be to rewrite [`memoryEfficientLoss`](https://github.com/pytorch/examples/blob/master/OpenNMT/train.py#L118), so that is first replicates the generator onto all used GPUs (say `n`), and then uses these modules with parallel_apply to process the `n` batches from the splits on each GPU. This could save a bit on replication, because that part is heavily CPU bound.

PR #81, opened 2017-02-23
add flush to print

fyu:
With flush, log info can appear immediately when it is directed to a pipe or file.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import argparse
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


apaszke:
This will break in Python 2, as `print` is not a function there 😕

fmassa:
@apaszke should we add `from __future__ import print_function`, or your prefer not to rely on that?

fyu:
print as a function is already used in master. This PR only adds an argument to the function. To make it compatible, we can have an additional PR to import print_function from __future__.

apaszke:
It's not used as a function. `print('asdf')` works both in Python 2 and 3. In 2 it just evaluates as a statement + an expression in parenthesis. We can merge it if you import the function.

fyu:
I see. thanks for the explanation. will do.

Maratyszcza:
This fails on Python 2.7: `Traceback (most recent call last):
  File "main.py", line 290, in <module>
    main()
  File "main.py", line 66, in main
    print("=> creating model '{}'".format(args.arch), flush=True)
TypeError: 'flush' is an invalid keyword argument for this function`

colesbury:
I'm going to revert this change. You can get the buffering behavior with either of the following:

```
unbuffer python main.py
python -u main.py
```

https://linux.die.net/man/1/unbuffer
https://docs.python.org/2/using/cmdline.html#cmdoption-u

PR #82, opened 2017-02-24
ONMT fixes and updates

bmccann:
Done:

- fixes bug in translate when user wishes to use cpu by passing -1 as device number

- removes re-initialization of checkpoint parameters

- replaces chunking with splitting

- removes unnecessary opt.cuda

- replace pre-ModuleList code with ModuleList

- converts to int when user wants to use the start_decay_at argument

-  fixes issue with decoder hidden state (https://github.com/pytorch/examples/issues/88)

- replaces from-scratch gradient clipping with nn.utils.clip_grad_norm

- adds additional logging for source and target tokens per second

- fixed: index was set to 0 in the translate.py code for printing the n best

- fixed: len was applied the batch (always giving batch size) rather than the top prediction for each batch when counting the total number of predicted words in translate.py

- added info on how one might go about prepping data for an MT task from scratch (Flickr30k) since demo data was already tokenized. 

- replace LSTMCell with a single-step single-layer LSTM (https://github.com/pytorch/examples/pull/80) [reverted to LSTMCell as pytorch overhead for calling cudnn makes the nn.LSTM version slower than the one based on LSTMCell]

- do not manually unroll encoder in onmt/Translator.py for bidirectional encoders. The manual unroll was broken for bidirectional encoders. The downside is that inference no longer masks off hidden states of bidirectional encoders for time steps corresponding to padding. My experiments show that this will affect BLEU scores if there is padding. For now, to avoid this issue, one can run with batch size 1 for models trained with -brnn. Once the code is updated with https://github.com/pytorch/pytorch/pull/873, this will no longer be an issue.

- "rewrite memoryEfficientLoss, so that it first replicates the generator onto all used GPUs (say n), and then uses these modules with parallel_apply to process the n batches from the splits on each GPU using parallel_apply" (https://github.com/pytorch/examples/pull/80) [Tried model parallelism for the generator and it provided no benefit over using data parallelism, so I ended up leaving the data parallelism]

### DIFF
--- a/OpenNMT/.gitignore
+++ b/OpenNMT/.gitignore
@@ -1,0 +1,3 @@
+pred.txt
+multi-bleu.perl
+*.pt
--- a/OpenNMT/README.md
+++ b/OpenNMT/README.md
@@ -32,19 +32,19 @@ Use of OpenNMT consists of four steps:
 
 ```perl multi-bleu.perl data/tgt-test.txt < demo_pred.txt```
 
-## WMT'16 Multimodal Translation: Flickr30k (de-en)
+## WMT'16 Multimodal Translation: Multi30k (de-en)
 
-Data might not come as clean as the demo data. Here is a second example that uses the Moses tokenizer (http://www.statmt.org/moses/) to prepare the Flickr30k data from the WMT'16 Multimodal Translation task (http://www.statmt.org/wmt16/multimodal-task.html).
+Data might not come as clean as the demo data. Here is a second example that uses the Moses tokenizer (http://www.statmt.org/moses/) to prepare the Multi30k data from the WMT'16 Multimodal Translation task (http://www.statmt.org/wmt16/multimodal-task.html).
 
 ### 0) Download the data.
 
-```mkdir -p data/flickr```
+```mkdir -p data/multi30k```
 
-```wget http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/training.tar.gz &&  tar -xf training.tar.gz -C data/flickr && rm training.tar.gz```
+```wget http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/training.tar.gz &&  tar -xf training.tar.gz -C data/multi30k && rm training.tar.gz```
 
-```wget http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz && tar -xf validation.tar.gz -C data/flickr && rm validation.tar.gz```
+```wget http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/validation.tar.gz && tar -xf validation.tar.gz -C data/multi30k && rm validation.tar.gz```
 
-```wget https://staff.fnwi.uva.nl/d.elliott/wmt16/mmt16_task1_test.tgz && tar -xf mmt16_task1_test.tgz -C data/flickr && rm mmt16_task1_test.tgz```
+```wget https://staff.fnwi.uva.nl/d.elliott/wmt16/mmt16_task1_test.tgz && tar -xf mmt16_task1_test.tgz -C data/multi30k && rm mmt16_task1_test.tgz```
 
 ### 1) Preprocess the data.
 
@@ -56,24 +56,23 @@ Data might not come as clean as the demo data. Here is a second example that use
 
 ```wget https://github.com/moses-smt/mosesdecoder/blob/master/scripts/share/nonbreaking_prefixes/nonbreaking_prefix.en```
 
-```for l in en de; do for f in data/flickr/*.$l; do if [[ "$f" != *"test"* ]]; then sed -i "$ d" $f; fi; perl tokenizer.perl -no-escape -l $l -q  < $f > $f.tok; done; done```
+```for l in en de; do for f in data/multi30k/*.$l; do if [[ "$f" != *"test"* ]]; then sed -i "$ d" $f; fi; perl tokenizer.perl -no-escape -l $l -q  < $f > $f.tok; done; done```
 
-```python preprocess.py -train_src data/flickr/train.en.tok -train_tgt data/flickr/train.de.tok -valid_src data/flickr/val.en.tok -valid_tgt data/flickr/val.de.tok -save_data data/flickr```
+```python preprocess.py -train_src data/multi30k/train.en.tok -train_tgt data/multi30k/train.de.tok -valid_src data/multi30k/val.en.tok -valid_tgt data/multi30k/val.de.tok -save_data data/multi30k```
 
 ### 2) Train the model.
 
-```python train.py -data data/flickr-train.pt -save_model flickr_model -gpus 0```
+```python train.py -data data/multi30k-train.pt -save_model multi30k_model -gpus 0```
 
 ### 3) Translate sentences.
 
-```python translate.py -gpu 0 -model flickr_model_e7_*.pt -src data/flickr/test.en.tok -tgt data/flickr/test.de.tok -replace_unk -verbose -output flickr_pred.txt```
->>>>>>> c87fc08... tips for non-demo mt via flickr30k example
+```python translate.py -gpu 0 -model multi30k_model_e13_*.pt -src data/multi30k/test.en.tok -tgt data/multi30k/test.de.tok -replace_unk -verbose -output multi30k_pred.txt```
 
 ### 4) Evaluate.
 
 ```wget https://raw.githubusercontent.com/moses-smt/mosesdecoder/master/scripts/generic/multi-bleu.perl```
 
-```perl multi-bleu.perl data/flickr/test.de < flickr_pred.txt```
+```perl multi-bleu.perl data/multi30k/test.de.tok < multi30k_pred.txt```
 
 ## Pretrained Models
 
--- a/OpenNMT/README.md
+++ b/OpenNMT/README.md
@@ -20,11 +20,12 @@ OpenNMT consists of three commands:
 
 2) Train the model.
 
-```python train.py -data data/demo-train.pt -save_model model -gpus 1```
+```python train.py -data data/demo-train.pt -save_model model -gpus 0```
 
 3) Translate sentences.
 
-```python translate.py -gpu 1 -model model_e13_*.pt -src data/src-test.txt -tgt data/tgt-test.txt -replace_unk -verbose```
+```python translate.py -gpu 0 -model model_e13_*.pt -src data/src-test.txt -tgt data/tgt-test.txt -replace_unk -verbose```
+>>>>>>> c19b7d3... README changes for multi-gpu
 
 ## Pretrained Models
 
--- a/OpenNMT/README.md
+++ b/OpenNMT/README.md
@@ -29,6 +29,7 @@ Use of OpenNMT consists of four steps:
 ### 4) Evaluate.
 
 ```wget https://raw.githubusercontent.com/moses-smt/mosesdecoder/master/scripts/generic/multi-bleu.perl```
+
 ```perl multi-bleu.perl data/tgt-test.txt < demo_pred.txt```
 
 ## WMT'16 Multimodal Translation: Flickr30k (de-en)
@@ -71,6 +72,7 @@ Data might not come as clean as the demo data. Here is a second example that use
 ### 4) Evaluate.
 
 ```wget https://raw.githubusercontent.com/moses-smt/mosesdecoder/master/scripts/generic/multi-bleu.perl```
+
 ```perl multi-bleu.perl data/flickr/test.de < flickr_pred.txt```
 
 ## Pretrained Models
--- a/OpenNMT/README.md
+++ b/OpenNMT/README.md
@@ -28,9 +28,10 @@ Use of OpenNMT consists of four steps:
 
 ### 4) Evaluate.
 
-```wget https://raw.githubusercontent.com/moses-smt/mosesdecoder/master/scripts/generic/multi-bleu.perl```
-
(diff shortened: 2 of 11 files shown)


apaszke:
Can you also replace [this part](https://github.com/pytorch/examples/blob/master/OpenNMT/onmt/Optim.py#L32-L42) with `torch.nn.utils.clip_grad_norm`?

bmccann:
@ngimel I've replaced the decoder's list of LSTMCells with a single multi-layer nn.LSTM. In your suggestions, you mentioned replacing each individual LSTMCell with an nn.LSTM, but shouldn't this be even faster? Instead, it appears to be 10% slower, and it runs out of memory. 

I'll slack you the profiler output in case you'd like to take a look.

bmccann:
After talking with @ngimel, thought I'd summarize some of what we are seeing:

Replacing LSTMCell with nn.LSTM and unrolling manually is indeed slower. This is due to overhead for weight copying. cuDNN is currently only faster when this overhead is amortized over many time steps.

Just to check, I tried replacing the cuDNN encoder with LSTMCell/StackedLSTM that the Decoder uses, and that was also slower, so it looks like the cuDNN encoder/LSTMCell Decoder combination is the sweet spot for the time being.

Running with the cudnn decoder also caused OOM errors with the same settings that the non-cudnn decoder could handle. @csarofeen investigated further and has reproduced a simple case here https://github.com/pytorch/pytorch/issues/914.

OpenNMT/train.py
```diff
@@ -258,11 +257,11 @@ def main():
...
             nn.LogSoftmax())
-        if len(opt.gpus) > 1:
-            generator = nn.DataParallel(generator, device_ids=opt.gpus)
+#        if len(opt.gpus) > 1:
+#            generator = nn.DataParallel(generator, device_ids=opt.gpus)
         model = onmt.Models.NMTModel(encoder, decoder, generator)
         if len(opt.gpus) > 1:
-            model = nn.DataParallel(model, device_ids=opt.gpus)
+            model = nn.DataParallel(model, device_ids=opt.gpus, dim=1)
         if opt.gpus:
```
ngimel: But that depends on #843 being merged (it is not merged yet)?

bmccann: Yeah, I'm hoping that functionality will be merged sooner rather than later. I'll give it a few days as I work on a better way to parallelize the generator. If it doesn't make it through by the time I finish, I guess I'll go back to dim=0 until it is merged.

csarofeen: #843 was just merged. Please, let me know if you find any problems with it.

bmccann:
Awesome! Thanks!

soumith:
As soon as you remove the [WIP] tag, i'll merge this in.

bmccann:
Implemented a first attempt at model parallelism for the generator, but it too is not speeding things up. I'm going to take another look tomorrow, but if it doesn't improve things by tomorrow evening, I'll remove the [WIP] and we can merge what's here.

bmccann:
Ran a bunch of experiments today testing out the model parallelism for the generator. It seems helpful only in the high batch size, large softmax, 4 or more GPU range. I'll post some details tomorrow and make sure that users can choose between data parallelism and model parallelism for the generator. Seems like we'll need real multiprocessing to make multi-GPU worth it for smaller datasets.

This change will require changing the saving procedure to using state_dicts, which is really something that should happen anyways, so I'll make that and the series of off-by-one errors I found in the code today my last updates. I think I'll have to do that tomorrow though. One more day should do it!

bmccann:
I think this is good to go. It would probably be a good idea if at least one person looked over the final draft.

OpenNMT/onmt/Models.py
```diff
@@ -98,39 +92,33 @@ def __init__(self, opt, dicts):
...
             outputs += [output]
 
         outputs = torch.stack(outputs)
-        return outputs.transpose(0, 1), h, attn
+        return outputs, hidden, attn
 
 
 class NMTModel(nn.Module):
 
-    def __init__(self, encoder, decoder, generator):
```
adamlerer: Can you use a default argument `generator=None` here and remove the `hasattr` below?

OpenNMT/onmt/Optim.py
```diff
@@ -29,19 +31,9 @@ def __init__(self, params, method, lr, max_grad_norm, lr_decay=1, start_decay_at
...
-
-        grad_norm = math.sqrt(grad_norm)
-        shrinkage = self.max_grad_norm / grad_norm
-
-        for param in self.params:
-            if shrinkage < 1:
-                param.grad.data.mul_(shrinkage)
-
+        if self.max_grad_norm:
+            clip_grad_norm(self.params, self.max_grad_norm)
```
adamlerer: You don't have to do this, but `Optim.py` is a pretty extraneous class, you could just move this little bit of logic to `train.py` and delete this.

OpenNMT/onmt/Translator.py
```diff
@@ -9,17 +10,34 @@ def __init__(self, opt):
...
         if opt.cuda:
-            self.model.cuda()
+            model.cuda()
+            generator.cuda()
         else:
-            self.model.cpu()
+            model.cpu()
+            generator.cpu()
+
+        model.generator = generator
```
adamlerer: Yeah, I think that `generator` should be a submodule of `NMTModule`, then you don't have to do all this extra work of separately moving it to cuda, setting it's state dict, etc.

OpenNMT/onmt/Translator.py
```diff
@@ -48,32 +66,38 @@ def buildTargetTokens(self, pred, src, attn):
...
-                encStates[1].data.index_fill_(1, batchPadIdx, 0)
-            context += [context_t]
+        encStates, context = None, None
 
-        encStates = (self.model._fix_enc_hidden(encStates[0]),
-                      self.model._fix_enc_hidden(encStates[1]))
+        if self.model.encoder.num_directions == 2:
+            # bidirectional encoder is negatively impacted by padding
+            # run with batch size 1 for improved translations
+            # This will be resolved when variable length LSTMs are used instead
```
adamlerer: Print this as a warning or just don't allow it, IIRC it behaves very badly when padding isn't dealt with. Although if you're planning to incorporate variable-length LSTMs immediately I guess it doesn't matter.

OpenNMT/train.py
```diff
@@ -61,22 +61,34 @@
...
 parser.add_argument('-curriculum', action="store_true",
                     help="""For this many epochs, order the minibatches based
                     on source sequence length. Sometimes setting this to 1 will
                     increase convergence speed.""")
+parser.add_argument('-extra_shuffle', action="store_true",
+                    help="""By default only shuffle mini-batch order; when true,
+                    shuffle and re-assign mini-batches""")
+
+#learning rate
+parser.add_argument('-fix_learning_rate', action='store_false', dest='update_learning_rate',
```
adamlerer: Doesn't -learning_rate go here?

adamlerer: This is a break with OpenNMT, and I think in general you don't want to decay learning rate for non-SGD optimizers, correct?

Also, you refer to "If update_learning_rate" below, but that's an internal parameter.

OpenNMT/train.py
```diff
@@ -61,22 +61,34 @@
...
+                    help="Do not decay learning rate (may be desirable for some optimzers (e.g. Adam)")
+parser.add_argument('-learning_rate_decay', type=float, default=0.5,
+                    help="""If update_learning_rate, decay learning rate by
+                    this much if (i) perplexity does not decrease on the
+                    validation set or (ii) epoch has gone past
+                    start_decay_at""")
+parser.add_argument('-start_decay_at', type=int, default=8,
```
adamlerer: Should make it clear that this is only for SGD (or only when -fix_learning_rate is not set).

OpenNMT/train.py
```diff
@@ -254,42 +287,47 @@ def main():
...
+    encoder = onmt.Models.Encoder(opt, dicts['src'])
+    decoder = onmt.Models.Decoder(opt, dicts['tgt'])
 
+    generator = nn.Sequential(
+        nn.Linear(opt.rnn_size, dicts['tgt'].size()),
+        nn.LogSoftmax())
+
+    model = onmt.Models.NMTModel(encoder, decoder)
+
+    if opt.train_from:
```
adamlerer: I guess I'm overruled about using `load_state_dict`. Notice that now you might accidently load the state from a model with different options (e.g. unidirectional vs. bidirectional), with undefined results (maybe some random error, maybe silently giving bad results).

Also, this means that everyone's checkpoints will be broken. You'll at least need to go fix the checkpoints mentioned in the README, sorry.

Also please use `if opt.train_from is not None`, it's safer in Python.

OpenNMT/onmt/Dataset.py
```diff
@@ -46,3 +50,9 @@ def __getitem__(self, index):
 
     def __len__(self):
         return self.numBatches
+
+
+    def shuffle(self):
```
bmccann: Yeah, I left it in as an option because I saw somewhat better scores using the completely shuffled data, though it does slow things down quite a bit. I can remove it.

OpenNMT/README.md
```diff
@@ -8,23 +8,69 @@ an open-source (MIT) neural machine translation system.
 
 ## Quickstart
 
-OpenNMT consists of three commands:
+Use of OpenNMT consists of four steps:
```
bmccann: Agreed

OpenNMT/.gitignore
```diff
@@ -0,0 +1,3 @@
+pred.txt
+multi-bleu.perl
+*.pt
```
bmccann: Will do

OpenNMT/onmt/Translator.py
bmccann: The main reason I didn't leave generator as a submodule was because both generator and model could be wrapped separately in DataParallel, and I wanted to be able to load state_dicts from models originally trained on X GPUs and continue training them on Y GPUs. Maybe having DataParallel in there wouldn't mess with the state_dicts the way I expected though. I'll double check.

OpenNMT/onmt/Translator.py
bmccann: Yeah, the brnn encoders will suffer quite a bit unless translate is run with batch size 1 at the moment. I'll add the warning. The unidirectional encoders were always fine.

OpenNMT/train.py
bmccann: Definitely should move -learning_rate. 

I can leave the -fix_learning_rate parameter out of the example, though I did find decay useful for some non-SGD optimizers. 

Ah, it was originally update_learning_rate, and I changed it to fix_learning_rate. I'll get rid of this all together if no one else has seen benefits from decaying non-SGD optimizers.

OpenNMT/train.py
bmccann: Yeah, seemed like state_dicts were being recommended in the docs. One of the benefits I see is in running things on multiple devices or switching between data parallelism and model parallelism quite easily. 

Okay, I'll look into those checkpoints.

Got it.

bmccann: Re: different model options. We could just load in the options from the checkpoint as well to ensure they are the same, though this gets in the way of things like manually lowering learning rates or changing batch sizes or device numbers between checkpoints.

bmccann: We could print out warnings for all the options that differ between the current run and the options saved in the checkpoint.

OpenNMT/onmt/Models.py
bmccann: I think I recall trying that and seeing that if I set generator to None in __init__, assigning a module to it later is a no-op -- just leaves it as None. I'll double check that.

OpenNMT/onmt/Models.py
```diff
@@ -98,39 +92,33 @@ def __init__(self, opt, dicts):
...
         outputs = []
         output = init_output
-        for i, emb_t in enumerate(emb.chunk(emb.size(0), dim=0)):
+        for emb_t in emb.split(1):
             emb_t = emb_t.squeeze(0)
             if self.input_feed:
                 emb_t = torch.cat([emb_t, output], 1)
 
-            output, h = self.rnn(emb_t, hidden)
+            output, hidden = self.rnn(emb_t, hidden)
```
bmccann: I think this might have somehow got introduced back when I was forking off your repo and things got mixed up in a rebase. My guess is that when you ran your benchmarks, it wasn't an issue. I'll check the 1.5x, but again, before the original multi-gpu PR went in, I think this was fine.

OpenNMT/onmt/Translator.py
adamlerer: Actually, better than this warning is to just raise and error if batch_size > 1 with brnn (and skip this whole padding logic altogether if batch_size=1).

OpenNMT/onmt/Models.py
adamlerer: Yeah, that doesn't seem right.
Anyway, the whole point of adding it as an optional arg is that you should assign it in the cosntructor instead of later on if possible.

OpenNMT/onmt/Models.py
adamlerer: Whew, that's good to know. I checked and indeed this bug was not in the original code.

OpenNMT/onmt/Models.py
bmccann: I thought it was weird too. I've opened https://github.com/pytorch/pytorch/issues/977 for now.

soumith:
there, i've merged it. if you have new fixes, do send them in a  new pr.
Thanks a lot for your work Bryan!

Issue #83, opened 2017-02-24
word_language_model: Some suggested arguments for `--model` are not valid


cuevasclemente:
In the word_language_model example, the suggestions for a `RNN_TANH` or `RNN_RELU` are not valid arguments. It looks like these haven't been valid members of the API since October 2016: https://github.com/pytorch/pytorch/commit/b5d13296c65e4b3cd5aa9715cf58df0fc043454e, instead the API determines the type of RNN based on the `mode` argument to the RNN class. 

An easy fix would be to check for either `RNN_TANH` or `RNN_RELU` and then pass that as the mode for the class, or use the `LSTM` or `GRU` models directly as is done currently in the code (by accessing the classes from `getattribute(nn, 'LSTM'/'GRU')`

PR #85, opened 2017-02-24
Command Line Interface backwards compatible fix for models.py

cuevasclemente:
To address https://github.com/pytorch/examples/issues/83

I did it in this way so that the command line interface to the program would not have to change (i.e: you could still pass `--model RNN_TANH` as described in the help for `model` in the argument parser), but if it is desirable to change that we can make the fix a little bit cleaner.

### DIFF
--- a/word_language_model/model.py
+++ b/word_language_model/model.py
@@ -12,14 +12,11 @@ class RNNModel(nn.Module):
             self.rnn = getattr(nn, rnn_type)(ninp, nhid, nlayers, bias=False)
         else:
             try:
-                nonlinearity = {
-                    'RNN_TANH': 'tanh', 'RNN_RELU': 'relu'}[rnn_type]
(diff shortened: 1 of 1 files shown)


word_language_model/model.py
```diff
@@ -8,7 +8,18 @@ class RNNModel(nn.Module):
     def __init__(self, rnn_type, ntoken, ninp, nhid, nlayers):
         super(RNNModel, self).__init__()
         self.encoder = nn.Embedding(ntoken, ninp)
-        self.rnn = getattr(nn, rnn_type)(ninp, nhid, nlayers, bias=False)
+        if rnn_type in ['LSTM', 'GRU']:
+            self.rnn = getattr(nn, rnn_type)(ninp, nhid, nlayers, bias=False)
+        else:
+            try:
+                nonlinearity = {
```
apaszke: Can you remove that newline? We accept up to 120 chars in a line.

apaszke:
Thanks!

PR #90, opened 2017-02-28
incorrect argument in README.md

donglixp:
There is no ``-cuda'' argument in train.py. Modify it to the correct one: -gpus 0 .


bmccann:
Yep, I forgot to change the README.md. https://github.com/pytorch/examples/pull/82 is going to fix it and the one a few lines down for translate.py.

PR #91, opened 2017-03-02
Update dcgan requirements and README

alykhantejani:
Added `lmdb` to the dcgan `requirements.txt` as the `LSUN` dataset needs this.

Also added some details to the dcgan README on how to download the LSUN dataset


soumith:
thanks!

PR #93, opened 2017-03-02
Use logging for messages in ImageNet example

Maratyszcza:
As a fix for #81 and #92, log messages using Python's built-in `logging` module. By default, the stream is flushed after every logged message.


soumith:
i think Sam's suggested is really sufficient. I dont see why we need to make simple code more complicated:
```
unbuffer python main.py
python -u main.py
```

PR #97, opened 2017-03-03
Update README.md

mrdrozdov:
Change flag from `-cuda` to `-gpu`.


apaszke:
That fix is already in #82.

PR #98, opened 2017-03-04
Update Models.py

ChunchuanLv:
should decoder rnn update hidden, instead of store it into a dummy h?

-            output, h = self.rnn(emb_t, hidden)
+            output, hidden = self.rnn(emb_t, hidden)


jekbradbury:
Fixed in #82.

PR #99, opened 2017-03-06
built-in optimizer to update parameters

Stonesjtu:
replace the reimplemented parameters updating method with the `SGD` optimizer.


soumith:
while this is nice. This is the only example that does explicit optimization rather than use `torch.optim`. I'm inclined to leave it as an illustration.

PR #100, opened 2017-03-06
Change the calculation method of the batch number. 

zihaofu:
Hi, I have changed the calculation method for batch number, since it will crash when dataset is smaller than 64. Since if dataset number is smaller than 64, `self.numBatches` will be 0 and will lead to crash.


soumith:
thanks!

PR #101, opened 2017-03-07
set the type of max_sent_length to int

sheng-z:
1. The type of `max_sent_length` should be set to **int**. 
2. In computing the `NLLLoss`, get the correct `batch_size`.


soumith:
OpenNMT has been moved to https://github.com/OpenNMT/OpenNMT-py and will be continuing development there by the OpenNMT team and other community members.

Please send pull requests over there.

Thank you.

PR #104, opened 2017-03-10
PTB LM example now has far better perplexity for both large (72.30) and small (87.17) models

Smerity:
This is as close to SotA as I can get from a standard LSTM without a form of recurrent dropout (see: https://github.com/pytorch/pytorch/issues/561).

Parameters are similar to [Recurrent Neural Network Regularization (Zaremba et al. 2014)](https://arxiv.org/pdf/1409.2329.pdf) and include the one line tweak from [Tying Word Vectors and Word Classifiers: A Loss Framework for Language Modeling (Inan et al. 2016)](https://arxiv.org/pdf/1611.01462.pdf) that substantially improves perplexity whilst minimizing the number of parameters.

This also serves as a better learning example now as well, given it includes dropout (including the use of `model.train()` and `model.eval()`) and also saves / reloads the model with best validation loss.

### DIFF
--- a/word_language_model/README.md
+++ b/word_language_model/README.md
@@ -5,12 +5,17 @@ By default, the training script uses the PTB dataset, provided.
 The trained model can then be used by the generate script to generate new text.
 
 ```bash
-python main.py --cuda  # Train an LSTM on ptb with cuda (cuDNN). Should reach perplexity of 113
-python generate.py     # Generate samples from the trained LSTM model.
+python main.py --cuda --epochs 6        # Train a LSTM on PTB with CUDA, reaching perplexity of 117.61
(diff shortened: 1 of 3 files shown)


soumith:
this is great, and looks good. Thanks!

HolyLow:
As discussed in [https://github.com/pytorch/examples/issues/415](url), many (including myself) fail to reproduce the claimed perplexity as reported in the README. I really want to know what I've done wrong. @soumith @Smerity

HolyLow:
It seems from [https://github.com/pytorch/examples/issues/281](url) that the perplexity in the README is based on PTB dataset while now the default provided dataset is WT2. Maybe the README should be updated.

soumith:
@HolyLow you are right, the README needs to be updated with baselines on WT2. For now, I am removing the perplexity values in the README

PR #108, opened 2017-03-13
Fix random seeding in DCGAN

DmitryUlyanov:
Added `manualSeed` argument and fixed cuda seeding. Also fixed some linting issues and typo.


soumith:
thanks!

PR #109, opened 2017-03-13
update attribution of weight tying in word_language_model

ofirpress:
As written in both Inan et. al. (2016) and Press & Wolf (2016), weight tying in RNN language models was developed in parallel by both of these teams, and so I've updated the attribution in the code to include both.


soumith:
i'll let @Smerity take a call here. I'm not familiar with relevant literature. Thanks.

Smerity:
Indeed, I should have cited both, thanks for fixing. LGTM, suggest merge.

soumith:
Thanks Ofir.

PR #114, opened 2017-03-19
Switch the model to evaluation mode before generation in the language model example

Henderake:
The model in the word language model example uses dropout, so I add a line to call model.eval() before generation.


soumith:
thanks!

PR #117, opened 2017-03-21
Remove rectifier before softmax

dmitriy-serdyuk:
As far as I am concerned, `softmax(relu(h_last))` is a very unusual thing to do. Removed the rectifier for convergence and didactic purposes as discussed [here](https://discuss.pytorch.org/t/relu-in-mnist-example/1190/2).


soumith:
thanks!

PR #118, opened 2017-03-22
Add a time sequence prediction example

maplewizard:
Hi,
I have made a time sequence prediction example, which I think, is capable to illustrate the usage of LSTMCell in pytorch. It's enough simple and fun and is a good small case for beginners who are interested in either pytorch or time sequence.

### DIFF
--- a/time_sequence_prediction/generate_sine_wave.py
+++ b/time_sequence_prediction/generate_sine_wave.py
@@ -1,0 +1,12 @@
+import math
+import numpy as np
+import cPickle as pickle
+T = 20
+L = 1000
+N = 100
+np.random.seed(2)
+x = np.empty((N, L), 'int64')
+x[:] = np.array(range(L)) + np.random.randint(-4*T, 4*T, N).reshape(N, 1)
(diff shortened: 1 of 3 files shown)


maplewizard:
@soumith , I have removed the OpenNMT code, and this is a new PR.

time_sequence_prediction/generate_sine_wave.py
```diff
@@ -0,0 +1,12 @@
+import math
+import numpy as np
+import cPickle as pickle
```
soumith: can you just use `torch.save` to dump these.

Also, import cPickle only works in python2 fyi

time_sequence_prediction/train.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import print_function
+import torch
+import torch.nn as nn 
+from torch.autograd import Variable
+import cPickle as pickle
```
soumith: remove cPickle and just use `torch.load`

maplewizard:
@soumith It has been replaced.

soumith:
thank you @maplewizard for your contribution.

PR #119, opened 2017-03-22
specify train and eval mode for netG and netD

iammarvelous:


soumith:
this is not really necessary. are you doing this in an attempt to save computation, or to get BatchNorm statistics right?

iammarvelous:
@soumith Yes. Maybe it won't influence the final result much. But it could corrupt BatchNorm statistics technically. And as a tutorial sample for new users to learn, it is good to remind them of using `train` and `eval` mode.

soumith:
your change in line 257 seems fine.
But your changes in other places seem inappropriate because the learning is very stochastic, and it is better to rely on batch statistics than running_mean / running_std in those places..

iammarvelous:
Well, it may not be the case. Since what we want are better generative results on `eval` mode rather than based on batch statistics, we should also treat the same for the input of D, right? Technically, I didn't see a problem there. But did you mean it may lead to some converge problems?

soumith:
i'm just saying that batch statistics are of better quality, so we should rather use those.
I'm not sure it leads to convergence problems though.

PR #121, opened 2017-03-24
fix a bug in DCGAN when using single gpu

huqinghao:
Using single GPU in DCGAN caused an error because  'gpu_ids'  in forward function will be None .


soumith:
thank you!

PR #124, opened 2017-03-27
language model: Handle tied + dimensions mismatch

nadavbh12:


Smerity:
Ah, good catch. LGTM.

soumith:
thanks!

PR #127, opened 2017-03-28
apply normalization for output image of dcgan generator

fehiepsi:
Fix the bug where generated images are not normalized in DCGAN example.


soumith:
ah yes of course. thanks for fixing it.

PR #129, opened 2017-03-29
fast-neural-style example

darkstar112358:
This is an implementation of an algorithm for artistic style transfer. The algorithm can be used to mix the content of an image with the style of another image. The implementation uses the method described in [Perceptual Losses for Real-Time Style Transfer and Super-Resolution](https://arxiv.org/abs/1603.08155) along with [Instance Normalization](https://arxiv.org/pdf/1607.08022.pdf).


darkstar112358:
hey @soumith , I tried using the pre-trained pytorch VGG models for this implementation. I can get good results, but the results look better if I use the Caffe pretrained VGG. The colors are more vivid when using the Caffe VGG as compared to pytorch VGG. 

Do you prefer using the pytorch VGG as compared to the Caffe VGG?

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+
+            xc = Variable(x.data.clone(), volatile=True)
+
+            y = utils.normalize_batch(y)
+            xc = utils.normalize_batch(xc)
+
+            features_y = vgg(y)
+            features_xc = vgg(xc)
+
+            f_xc_c = Variable(features_xc[1].data, requires_grad=False)
```
fmassa: A small comment in here: I think it would be great for readability if we could have `features_xc` be a `namedtuple`, as it's not really clear from which layer you are taking this information from.

fast_neural_style/neural_style/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+def gram_matrix(y):
+    (b, ch, h, w) = y.size()
+    features = y.view(b, ch, w * h)
+    features_t = features.transpose(1, 2)
+    gram = features.bmm(features_t) / (ch * h * w)
+    return gram
+
+
+def normalize_batch(batch):
```
fmassa: Can't you use `torchvision.transforms.Normalize` in here? There is the 255 normalization that differs (but I think it could be removed as in my previous comment`, but even if we keep it, it could just be added as a LambdaTransform, right?

fast_neural_style/neural_style/transformer_net.py
```diff
@@ -0,0 +1,140 @@
...
+        y = self.relu(self.in4(self.deconv1(y)))
+        y = self.relu(self.in5(self.deconv2(y)))
+        y = self.deconv3(y)
+        return y
+
+
+class ConvLayer(torch.nn.Module):
+    def __init__(self, in_channels, out_channels, kernel_size, stride):
+        super(ConvLayer, self).__init__()
+        reflection_padding = int(np.floor(kernel_size / 2))
```
fmassa: I think this is equivalent to `kernel_size // 2`

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+    train_loader = DataLoader(train_dataset, batch_size=args.batch_size)
+
+    transformer = TransformerNet()
+    optimizer = Adam(transformer.parameters(), args.lr)
+    mse_loss = torch.nn.MSELoss()
+
+    vgg = Vgg16()
+    style_transform = transforms.Compose([
+        transforms.ToTensor(),
+        transforms.Lambda(lambda x: x.mul(255))
```
fmassa: Same as before, do we need this multiplication? It might require changing the training hyperparameters though, as the scaling for the MSE loss will be multiplied by a factor.

fast_neural_style/neural_style/utils.py
fmassa: Just saw that tis works in 4D tensors, while normalize works on 3D tensors only.
But some parts could be simplified. For example, you can call `tensor.new(size)` to get a new tensor of the same type (and device) as `tensor`, of size `size`

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+    torch.manual_seed(args.seed)
+
+    if args.cuda:
+        torch.cuda.manual_seed(args.seed)
+
+    transform = transforms.Compose([
+        transforms.Scale(args.image_size),
+        transforms.CenterCrop(args.image_size),
+        transforms.ToTensor(),
+        transforms.Lambda(lambda x: x.mul(255))
```
darkstar112358: In my experience the current configuration works better than the normalizing at the beginning. The generator gives better results if we let it generate image with pixels in range [0, 255]. The other option is to let the generator produce image in [0, 1] range and we denormalize it before saving to file but this method gives poorer results. I tried tuning the parameter and definitely got better results with the current method.

The multiplication by 255 is needed because we need to compare the image produced by the generator with our training set images.

fast_neural_style/neural_style/neural_style.py
darkstar112358: Discussed above.

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+                x = x.cuda()
+
+            y = transformer(x)
+
+            xc = Variable(x.data.clone(), volatile=True)
+
+            y = utils.normalize_batch(y)
+            xc = utils.normalize_batch(xc)
+
+            features_y = vgg(y)
```
darkstar112358: The weights of VGG are not to be updated. They were not being updated earlier because the optimizer used only the parameters of generator. I have detached the parameters of VGG as you suggested, [code](https://github.com/darkstar112358/examples/blob/24bfb3c1ec16868eb336db18a83b17fd022ea57c/fast_neural_style/neural_style/vgg.py#L25-L27). I don't know of a better way of not to require gradients.

fast_neural_style/neural_style/utils.py
darkstar112358: I modified the code to use `tensor.new(size)` way. As you noticed this has to be done on 4D tensors.

darkstar112358:
Thanks for the review @fmassa 😃. I have incorporated the changes you suggested in the code except the 255 comment. I have discussed that issue above.

Thanks

fast_neural_style/neural_style/neural_style.py
darkstar112358: Fixed in 24bfb3c1ec16868eb336db18a83b17fd022ea57c.

fast_neural_style/neural_style/transformer_net.py
darkstar112358: Fixed in 24bfb3c1ec16868eb336db18a83b17fd022ea57c.

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+
+    if args.cuda:
+        transformer.cuda()
+        vgg.cuda()
+        style = style.cuda()
+
+    style_v = Variable(style, volatile=True)
+    style_v = utils.normalize_batch(style_v)
+    features_style = vgg(style_v)
+    gram_style = [utils.gram_matrix(y) for y in features_style]
```
darkstar112358: Fixed in 24bfb3c1ec16868eb336db18a83b17fd022ea57c.

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+
+            f_xc_c = Variable(features_xc[1].data, requires_grad=False)
+
+            content_loss = args.content_weight * mse_loss(features_y[1], f_xc_c)
+
+            style_loss = 0.
+            for m in range(len(features_y)):
+                gram_s = Variable(gram_style[m].data, requires_grad=False)
+                gram_y = utils.gram_matrix(features_y[m])
+                style_loss += args.style_weight * mse_loss(gram_y, gram_s[:n_batch, :, :])
```
darkstar112358: Fixed in 24bfb3c1ec16868eb336db18a83b17fd022ea57c.

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,231 @@
...
+
+            features_y = vgg(y)
+            features_xc = vgg(xc)
+
+            f_xc_c = Variable(features_xc[1].data, requires_grad=False)
+
+            content_loss = args.content_weight * mse_loss(features_y[1], f_xc_c)
+
+            style_loss = 0.
+            for m in range(len(features_y)):
```
darkstar112358: Fixed in 24bfb3c1ec16868eb336db18a83b17fd022ea57c.

fast_neural_style/neural_style/vgg.py
```diff
@@ -0,0 +1,40 @@
...
+        for x in range(4, 9):
+            self.slice2.add_module(str(x), vgg_pretrained_features[x])
+        for x in range(9, 16):
+            self.slice3.add_module(str(x), vgg_pretrained_features[x])
+        for x in range(16, 23):
+            self.slice4.add_module(str(x), vgg_pretrained_features[x])
+
+        if detach_parameters:
+            for param in self.parameters():
+                param.detach()
```
fmassa: This doesn't sound right to me. Can't you just do something like
```python
for param in self.parameters():
    param.requires_grad = False
```

fast_neural_style/neural_style/transformer_net.py
```diff
@@ -0,0 +1,138 @@
...
+    def _reset_parameters(self):
+        self.scale.data.uniform_()
+        self.shift.data.zero_()
+
+    def forward(self, x):
+        n = x.size(2) * x.size(3)
+        t = x.view(x.size(0), x.size(1), n)
+        mean = torch.mean(t, 2).unsqueeze(2).expand_as(x)
+        # Calculate the biased var. torch.var returns unbiased var
+        var = torch.var(t, 2).unsqueeze(2).expand_as(x) * ((n - 1) / float(n))
```
fmassa: There is an option to compute the biased variance in torch (and in [PyTorch autograd](https://github.com/pytorch/pytorch/blob/master/torch/autograd/variable.py#L489) ), not sure why it's not wrapped in `pytorch`.
You can call it here via `torch.var(t, dim=2, unbiased=False)`

fmassa: I opened an issue for that https://github.com/pytorch/pytorch/issues/1412

fast_neural_style/neural_style/transformer_net.py
```diff
@@ -0,0 +1,138 @@
...
+    def forward(self, x):
+        x_in = x
+        if self.upsample:
+            x_in = self.upsample_layer(x_in)
+        out = self.reflection_pad(x_in)
+        out = self.conv2d(out)
+        return out
+
+
+class InstanceNormalization(torch.nn.Module):
```
fmassa: Actually, [`InstanceNormalization` was recently merged into master](https://github.com/pytorch/pytorch/pull/1283), and should be more efficient than this implementation because it leverages C / Cuda code directly. Maybe we should use that implementation here?

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,232 @@
...
+            n_batch = len(x)
+            count += n_batch
+            optimizer.zero_grad()
+            x = Variable(x)
+            if args.cuda:
+                x = x.cuda()
+
+            y = transformer(x)
+
+            xc = Variable(x.data.clone(), volatile=True)
```
fmassa: I think that if you make all weights of the `vgg` network `requires_grad=False`, then it's not necessary to have inputs being `volatile` to have low memory usage. See [discussion here](https://pytorch.slackarchive.io/random/page-9/ts-1490356747249255). It would remove the need of always unwrapping / wrapping the variables for the losses, and would make the code much cleaner I think.

fast_neural_style/neural_style/neural_style.py
```diff
@@ -0,0 +1,232 @@
...
+                                  help="size of training images, default is 256 X 256")
+    train_arg_parser.add_argument("--style-size", type=int, default=None,
+                                  help="size of style-image, default is the original size of style image")
+    train_arg_parser.add_argument("--cuda", type=int, required=True,
+                                  help="set it to 1 for running on GPU, 0 for CPU")
+    train_arg_parser.add_argument("--seed", type=int, default=42,
+                                  help="random seed for training")
+    train_arg_parser.add_argument("--content-weight", type=float, default=1e5,
+                                  help="weight for content-loss, default is 1e5")
+    train_arg_parser.add_argument("--style-weight", type=float, default=1e10,
```
fmassa: The values for the `content-weight` and `style-weight` looks very large. Is this needed? Checking some other implementations, they [have much more normalized values](https://github.com/jcjohnson/fast-neural-style/blob/master/train.lua#L36)

fast_neural_style/neural_style/vgg.py
darkstar112358: Incorporated your suggestion [[code](https://github.com/darkstar112358/examples/blob/fast-neural-style/fast_neural_style/neural_style/vgg.py#L24-L25)].

fast_neural_style/neural_style/transformer_net.py
darkstar112358: Thanks, I am using the inbuilt `InstanceNorm2d` so calculating the `var` is not needed.

fast_neural_style/neural_style/transformer_net.py
darkstar112358: Using inbuilt `InstanceNorm2d` now.

fast_neural_style/neural_style/neural_style.py
darkstar112358: Refactored the code according to this, code comes out cleaner.

fast_neural_style/neural_style/neural_style.py
darkstar112358: When I was using the pretrained VGG Caffe model with RGB values in the range `[0-255] ` I had the `content_weight` and `style_weight` on the same scale as the one listed in the [original paper](http://cs.stanford.edu/people/jcjohns/eccv16/) and other implementations, but when I started using pytorch's pretrained vgg which had RGB values in the range `[0-1]` I had to increase the scale of the cost lambdas.

I discussed this with a friend who also had to change the scale of cost lambdas when he was using the pretrained VGG for Keras(theano-backend).

darkstar112358:
Thanks for the review @fmassa, I will have to retrain the 4 example models to use the latest `InstanceNorm2d` code. I will update the models and get back to you once I have finished training them.

darkstar112358:
Hi @fmassa, I have incorporated all your review comments and I have retrained and updated the saved models with the latest code which uses the inbuilt `InstanceNorm2d` code. Regarding the `content-weight` and `style-weight` I have commented above.

soumith:
Thanks a lot @darkstar112358 
This is much needed.

abhiskk:
Thanks @soumith , happy to make a contribution!

exiaxu:
Can you explain why the hyperparameter `args.content_loss` and `args.style_loss` is so huge while in [Justin Johnson's code](https://github.com/jcjohnson/fast-neural-style/blob/master/train.lua) these parameter is much smaller ?

abhiskk:
@exiaxu this is because of different preprocessing needed for the pytorch pretrained vgg. I have a version of the algorithm [here](https://github.com/abhiskk/fast-neural-style) which uses the pretrained caffe vgg and has the same `args.content_loss` and `args.style_loss` scales.

MLnick:
@abhiskk hey there, was just wondering if there is an explicit license on the pre-trained weights you have at https://github.com/pytorch/examples/tree/master/fast_neural_style#models? (link to dropbox `https://www.dropbox.com/s/lrvwfehqdcxoza8/saved_models.zip?dl=0`). Are they freely available to use?

MLnick:
@abhiskk thanks very much!

PR #131, opened 2017-03-30
Fix the single gpu bug

wsstriving:
Original code will report a "gupid equals None error" when training on single gpu.


dcgan/main.py
```diff
@@ -131,7 +131,9 @@ def forward(self, input):
         gpu_ids = None
         if isinstance(input.data, torch.cuda.FloatTensor) and self.ngpu > 1:
             gpu_ids = range(self.ngpu)
-        return nn.parallel.data_parallel(self.main, input, gpu_ids)
+            return nn.parallel.data_parallel(self.main, input, gpu_ids)
+        else:
+            return self.main.forward(input)
```
soumith: this has to be `self.main(input)`

(and same below)

wsstriving: Fixed that

soumith:
fixed in master

PR #133, opened 2017-04-07
Fix Python 3 issue with time_sequence_prediction

dnouri:
open without 'rb' caused Python 3 to open this in text mode and fail


soumith:
thank you!

PR #136, opened 2017-04-10
face-recognition-with-center-loss example

BestSonny:
This is an implementation of an algorithm for face recognition with center loss. The algorithm can be used to minimize the intraclass distances of the deep features. The implementation uses the method [A Discriminative Feature Learning Approach
for Deep Face Recognition](http://ydwen.github.io/papers/WenECCV16.pdf)

### DIFF
--- a/center_loss/face_model.py
+++ b/center_loss/face_model.py
@@ -1,0 +1,36 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.multiprocessing as mp
+import torch.utils.model_zoo as model_zoo
+from torchvision.models.resnet import BasicBlock, ResNet, model_urls
+import math
+
+class FaceModel(nn.Module):
+    def __init__(self,num_classes, pretrained=False, **kwargs):
+        super(FaceModel, self).__init__()
+        self.model = ResNet(BasicBlock, [3, 4, 6, 3], **kwargs)
+        if pretrained:
+            parameters =  model_zoo.load_url(model_urls['resnet50'])
+            self.model.load_state_dict(parameters)
+        self.model.avgpool = None
+        self.model.fc = nn.Linear(512*3*4, 512)
+        self.model.classifier = nn.Linear(512, num_classes)
+        self.centers = torch.zeros(num_classes, 512).type(torch.FloatTensor)
+        self.num_classes = num_classes
+
+    def forward(self, x):
+        x = self.model.conv1(x)
+        x = self.model.bn1(x)
+        x = self.model.relu(x)
+        x = self.model.maxpool(x)
+        x = self.model.layer1(x)
+        x = self.model.layer2(x)
+        x = self.model.layer3(x)
+        x = self.model.layer4(x)
+        x = x.view(x.size(0), -1)
+        x = self.model.fc(x)
+        #feature for center loss
+        self.features = x
+        x = self.model.classifier(x)
+        return x
--- a/center_loss/main.py
+++ b/center_loss/main.py
@@ -12,9 +12,9 @@ import os
 
 # Training settings
 parser = argparse.ArgumentParser(description='PyTorch face recognition Example')
-parser.add_argument('--batch-size', type=int, default=128, metavar='N',
+parser.add_argument('--batch_size', type=int, default=128, metavar='N',
                     help='input batch size for training (default: 128)')
-parser.add_argument('--test-batch-size', type=int, default=64, metavar='N',
+parser.add_argument('--test_batch_size', type=int, default=64, metavar='N',
                     help='input batch size for testing (default: 64)')
 parser.add_argument('--epochs', type=int, default=10, metavar='N',
                     help='number of epochs to train (default: 10)')
@@ -24,6 +24,7 @@ parser.add_argument('--max_iter', type=int, default=100000, metavar='S',
                     help='random seed (default: 1)')
 parser.add_argument('--lr', type=float, default=0.001, help='learning rate, default=0.001')
 parser.add_argument('--beta1', type=float, default=0.5, help='beta1 for adam. default=0.5')
+parser.add_argument('--center_loss_weight', type=float, default=0.01, help='weight for center loss')
 parser.add_argument('--root', type=str,
         help='path to the data directory containing aligned face patches. Multiple directories are separated with colon.',
         default='./datasets/raw')
@@ -85,6 +86,7 @@ def main():
         train_loader=train_loader,
         val_loader=val_loader,
         max_iter=args.max_iter,
+        center_loss_weight = args.center_loss_weight
     )
 
     if args.resume:
--- a/center_loss/main.py
+++ b/center_loss/main.py
@@ -1,0 +1,114 @@
+from __future__ import print_function
+import argparse
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torchvision.transforms as transforms
+from face_model import FaceModel
+from lfw import LFW
+import trainer as tn
+import math
+import os
+
+# Training settings
+parser = argparse.ArgumentParser(description='PyTorch face recognition Example')
+parser.add_argument('--batch-size', type=int, default=128, metavar='N',
+                    help='input batch size for training (default: 128)')
+parser.add_argument('--test-batch-size', type=int, default=64, metavar='N',
+                    help='input batch size for testing (default: 64)')
+parser.add_argument('--epochs', type=int, default=10, metavar='N',
+                    help='number of epochs to train (default: 10)')
+parser.add_argument('--seed', type=int, default=1, metavar='S',
+                    help='random seed (default: 1)')
+parser.add_argument('--max_iter', type=int, default=100000, metavar='S',
+                    help='random seed (default: 1)')
+parser.add_argument('--lr', type=float, default=0.001, help='learning rate, default=0.001')
+parser.add_argument('--beta1', type=float, default=0.5, help='beta1 for adam. default=0.5')
+parser.add_argument('--root', type=str,
+        help='path to the data directory containing aligned face patches. Multiple directories are separated with colon.',
+        default='./datasets/raw')
+parser.add_argument('--resume', type=str,
+        help='model path to the resume training',
+        default='')
+
+# custom weights initialization
+def weights_init(m):
+    classname = m.__class__.__name__
+    if classname.find('Conv') != -1:
+        m.weight.data.normal_(0.0, 0.02)
+        n = m.kernel_size[0] * m.kernel_size[1] * m.out_channels
+        m.weight.data.normal_(0, math.sqrt(2. / n))
+        if m.bias is not None:
+            m.bias.data.zero_()
+    elif classname.find('BatchNorm') != -1:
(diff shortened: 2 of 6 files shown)


liorshk:
Why are you training and testing on the same directory?

BestSonny:
@liorshk Because in the LFW dataset, each class only has several pictures. This demo code only presents a brief description on how to train and validate a classification network with the center loss.

liorshk:
But it causes overfitting and therefore doesn't really validate the learning process. I think the example should be trained on some other dataset and tested on LFW.
The other option would be to split LFW to train/test sets.
Either way the number of classes in the last layer should be changed based on the dataset.

center_loss/data.py
```diff
@@ -0,0 +1,108 @@
...
+import os
+import os.path
+import errno
+import torch
+import json
+import codecs
+import numpy as np
+import random
+
+class ImageClass():
```
fmassa: Can you make it inherit from `object` so that it's a new-style class in Python 2?

center_loss/data.py
```diff
@@ -0,0 +1,108 @@
...
+        classes = os.listdir(path_exp)
+        classes.sort()
+        nrof_classes = len(classes)
+        for i in range(1000):
+            class_name = classes[i]
+            facedir = os.path.join(path_exp, class_name)
+            if os.path.isdir(facedir):
+                images = os.listdir(facedir)
+                image_paths = [os.path.join(facedir,img) for img in images]
+                dataset.append(ImageClass(class_name, image_paths))
```
fmassa: Do we need a class for that? Couldn't we just use a tuple instead?

center_loss/face_model.py
```diff
@@ -0,0 +1,61 @@
...
+import torch.nn.functional as F
+import torch.multiprocessing as mp
+import torch.utils.model_zoo as model_zoo
+from torchvision.models.resnet import BasicBlock, ResNet, model_urls
+import math
+
+class FaceModel(nn.Module):
+    def __init__(self,num_classes, pretrained=False, **kwargs):
+        super(FaceModel, self).__init__()
+        self.model = ResNet(BasicBlock, [2, 2, 2, 2], **kwargs)
```
fmassa: This is resnet18, but in the `pretrained` condition you load `resnet50`. Could you fix it?

center_loss/face_model.py
```diff
@@ -0,0 +1,61 @@
...
+        if pretrained:
+            parameters =  model_zoo.load_url(model_urls['resnet50'])
+            self.model.load_state_dict(parameters)
+        self.model.avgpool = None
+        self.model.fc1 = nn.Linear(512*3*4, 512)
+        self.model.fc2 = nn.Linear(512, 512)
+        self.model.classifier = nn.Linear(512, num_classes)
+        self.centers = torch.zeros(num_classes, 512).type(torch.FloatTensor)
```
fmassa: You could define it as a `buffer` via `register_buffer`. This way, when you call `.cuda()`, `centers` is automatically sent to the GPU. Also, it's automatically loaded when you call `load_state_dict()`, and is also present in `state_dict()`.

center_loss/face_model.py
```diff
@@ -0,0 +1,61 @@
...
+        x = self.model.maxpool(x)
+        x = self.model.layer1(x)
+        x = self.model.layer2(x)
+        x = self.model.layer3(x)
+        x = self.model.layer4(x)
+        x = x.view(x.size(0), -1)
+        x = self.model.fc1(x)
+        #feature for center loss
+        x = self.model.fc2(x)
+        self.features = x
```
fmassa: I'm not sure it's a good idea to save features in the model. Do you really need to save it?
Can't you just return it as a second output in your model?

center_loss/main.py
```diff
@@ -0,0 +1,132 @@
...
+parser.add_argument('--center_loss_weight', type=float, default=0.1, help='weight for center loss')
+parser.add_argument('--root', type=str,
+        help='path to the data directory containing aligned face patches. Multiple directories are separated with colon.',
+        default='./datasets/casia_mtcnnpy_160')
+parser.add_argument('--resume', type=str,
+        help='model path to the resume training',
+        default='')
+
+# custom weights initialization
+def weights_init(m):
```
fmassa: Can you use `torch.nn.init` instead?

center_loss/data.py
```diff
@@ -0,0 +1,108 @@
...
+        if len(self.dataset) == 0:
+            raise(RuntimeError("Found 0 images in subfolders of: " + self.root + "\n"
+                               "Supported image extensions are: " + ",".join(IMG_EXTENSIONS)))
+        self.transform = transform
+        self.target_transform = target_transform
+        self.loader = loader
+
+
+    def __getitem__(self, index):
+        if self.train:
```
fmassa: I think it would be cleaner to have the `if train else` logic in the constructor of the class, instead of having it selectively chose during `__getitem__`.

center_loss/data.py
```diff
@@ -0,0 +1,108 @@
...
+        return len(self.image_paths)
+
+def get_dataset(paths):
+    dataset = []
+    for path in paths.split(':'):
+        path_exp = os.path.expanduser(path)
+        classes = os.listdir(path_exp)
+        classes.sort()
+        nrof_classes = len(classes)
+        for i in range(1000):
```
fmassa: Do you need to hard-code it?

center_loss/main.py
```diff
@@ -0,0 +1,132 @@
...
+                                          transforms.ToTensor(),
+                                          transforms.Normalize([0.5,0.5,0.5], [0.5,0.5,0.5])])
+    train_dataset = ImageDataset(root, transform=train_transforms)
+    val_dataset = copy.deepcopy(train_dataset)
+    val_dataset.train = False
+
+    train_loader = torch.utils.data.DataLoader(train_dataset,batch_size=args.batch_size, shuffle=True, **kwargs)
+    val_loader = torch.utils.data.DataLoader(val_dataset,batch_size=args.test_batch_size, shuffle=False, **kwargs)
+
+    # train_loader = torch.utils.data.DataLoader(
```
fmassa: Can you clean up the comments?

center_loss/trainer.py
```diff
@@ -0,0 +1,174 @@
...
+        self.top1 = AverageMeter()
+        self.losses = AverageMeter()
+        self.best_prec1 = 0
+        self.center_loss_weight =  center_loss_weight
+        self.alpha = alpha
+
+    def validate(self):
+        self.model.eval()
+        top1 = AverageMeter()
+        losses = AverageMeter()
```
fmassa: Do you use it anywhere?

center_loss/trainer.py
```diff
@@ -0,0 +1,174 @@
...
+            if self.cuda:
+                data, target = data.cuda(), target.cuda(async=True)
+                self.model.centers = self.model.centers.cuda()
+
+            data_var = Variable(data, volatile=True)
+            target_var = Variable(target, volatile=True)
+            target_var = Variable(target, volatile=True)
+            # compute output
+            output = self.model(data_var)
+            loss = self.criterion(output, target_var)
```
fmassa: If you won't print the loss at some point, no need to compute it.

BestSonny:
@fmassa  Thanks for the precious comments and I learn a lot from them. I made some modification according to your feedback. 

@liorshk  I created a split function for splitting the training and the testing. https://github.com/BestSonny/examples/blob/7bcf8e1c83fdfcff1d97a1d2303276ca2d933d4a/center_loss/data.py#L51-#L65

liorshk:
@BestSonny Splitting to train and test is good but actually in the original paper they tested on pairs of images (LFW dataset) by checking the distance between the feature vectors (the output before the last layer). Can you report your results?

BestSonny:
@liorshk I didn't do the post-process testing yet. It takes a long time to train the network. Instead, I can find a time to run the MNIST experiment as in the original paper.

center_loss/face_model.py
bertjiazheng: When I access `model._buffer` after `model = torch.nn.DataParallel(model).cuda()`, `keyError` occurs. So how can I access `model._buffer` ?

log26:
I add center loss to resnet50. Although I set the batch size to 1, it's still out of memory when backward. Is it reasonable?

BestSonny:
@log26 I can not figure out your issues without more details.

log26:
@BestSonny  I use resnet50 to do classification task, and add center loss after avgpool layer (the dimension of features is 2048).  When the network backward, the memory requires more than 12G even I set the batch size to 1.  Is it reasonable? Or maybe I do something wrong?

BestSonny:
@log26 I didn't see such a problem. Can you print out your log or model?

log26:
image

log26:
@BestSonny   Here is my model
image
image
image

image

zhouyong64:
@BestSonny The code works great. However, it seems that it doesn't work with multi-GPU with nn.DataParallel.

center_loss/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get_dataset(paths):
+    dataset = []
+    for path in paths.split(':'):
+        path_exp = os.path.expanduser(path)
+        classes = os.listdir(path_exp)
+        classes.sort()
+        nrof_classes = len(classes)
+        for i in range(nrof_classes):
+            class_name = classes[i]
```
aivision2020: simply do for class_name in classes

center_loss/data.py
aivision2020: +1
you should init 2 instances of ImageDataset. one for train, one for val

center_loss/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+class ImageDataset(data.Dataset):
+    def __init__(self, root, transform=None, target_transform=None,
+                 train=True, split=0.1, loader=default_loader):
+        self.root = root
+        self.train = train
+        self.split = split
+        self.dataset, self.classes = get_dataset(self.root)
+        image_paths, labels = get_image_paths_and_labels(self.dataset)
+        self.train_image_paths, self.train_labels, self.val_image_paths, self.val_labels = split_train_test(image_paths, labels, self.split)
```
louis-she: In LFW dataset, some of the classes(persons) have only 1 image. Is it not meaningful to put these kind of data as validation dataset?

center_loss/face_model.py
VishalLatiyan: How do I divide LFW dataset into test/train? Some class only have 1 image.

louis-she: The author put it only in training set according to the 2 lines:
```
val_indices.extend(indexes[:int(np.floor(len(indexes)*split))])
train_indices.extend(indexes[int(np.floor(len(indexes)*split)):])
```
In my code, I put it both on validation and training set. I'm also not sure if it's ok to do that. You can reach the code at https://github.com/louis-she/center-loss.pytorch/blob/master/dataset.py#L65

louis-she: Just found out that one should really use `pairsDevTest.txt` to get the evaluation pairs. Suggested by FLW official. Refer to http://vis-www.cs.umass.edu/lfw/README.txt

PR #140, opened 2017-04-11
Changes in `reinforce.py`

zuoxingdong:


soumith:
there seem to be additional changes that are not just a typo

zuoxingdong:
@soumith Default value of `random seed` was not consistent with helper message, `saved_actions` are not used and `running_reward` should be initialized to be 0, if I got it correctly.

zuoxingdong:
@soumith  By the way, I am quite curious to see an additional example of `Vanilla Policy Gradient`, with advantage estimate (subtract baseline to reduce variance).

jekbradbury:
`actor_critic.py` uses an advantage estimate (subtracts the output of a value network) to reduce variance. And I believe `running_reward` should not be initialized to zero but rather to approximately the typical early value of the reward.

zuoxingdong:
@jekbradbury Thanks for the info ! I am a bit confused if `actor_critic.py` actually implements the Vanilla Policy Gradient algorithm, e.g. in [John Schulman's lecture note ](http://rll.berkeley.edu/deeprlcourse/docs/lec2.pdf) slide 17 ?

jekbradbury:
Slide 17 describes policy gradient with a scalar baseline that's updated with least-squares; `actor_critic.py` uses a value function baseline with an infinite time horizon, as described on slides 20-22. The terminology is definitely confusing -- many people would say policy gradient with a value function baseline is not "true" actor-critic, but I believe it's exactly the algorithm referred to as "advantage actor-critic" in the A3C paper...

zuoxingdong:
@jekbradbury Great explanation, thanks a lot. By the way, [This line](https://github.com/pytorch/examples/blob/master/reinforcement_learning/reinforce.py#L66) centralize the rewards, just to confirm if I understand it right that it is additional to the original algorithm ?

By the way, I have left put `running_reward` unchanged, i.e. `10`.

soumith:
thanks for the PR!

PR #141, opened 2017-04-12
save/load optimizer state

aromnvidia:
ImageNet example doesn't save optimizer state at the checkpoint. I added save and load instructions.


soumith:
thanks!

PR #145, opened 2017-04-18
snli/model.py: can run using Python 2.*

donglixp:
Python 2.* does not support ``out.view(*size, -1)''. Expand it to ``out.view(size[0], size[1], -1)''.

  File "train.py", line 9, in <module>
    from model import SNLIClassifier
  File "snli/model.py", line 13
    return out.view(*size, -1)
SyntaxError: only named arguments may follow *expression


soumith:
thanks!

PR #149, opened 2017-04-22
Update main.py

zsong:


zsong:
model simply doesn't take any variables.

soumith:
model's call operator should be called, not `.forward`

PR #150, opened 2017-04-23
Change reusing of Variables in DCGAN

bartolsthoorn:
This should not change the performance of DCGAN but it is perhaps good to teach people to not reuse Variables.


soumith:
thanks!

PR #153, opened 2017-04-26
Fix help message for no-cuda

jucor:
Change the help message for `--no-cuda`, explaining it *disables* CUDA, not *enables* it as currently written. The joy of boolean optional command line arguments :)


soumith:
thanks!

PR #154, opened 2017-04-28
replace model by policy

quanvuong:
It seems to me that `policy` is a better terminology choice than `model`. I was a bit confused when `model` was introduced and I hope this change removes the confusion for future readers. I'm using David Silver's definition:

Policy: agent’s behaviour function
Model: agent’s representation of the environment

But if I'm misunderstanding anything, I'll be very appreciative of any explanation. Thanks!


soumith:
thank you. makes sense.

PR #160, opened 2017-05-19
mnist_hogwild manual breaking of gradient sharing removed

pfre00:
Since gradients are allocated lazily, they are instantiated in every process separately on the `loss.backward()` call. The removed lines in train.py are no longer needed.


apaszke:
Thank you!

PR #175, opened 2017-07-06
Bugs In MNIST

chao1224:
Found one bug in MNIST.
1. Go to this line: [MNIST main script](https://github.com/pytorch/examples/blob/master/mnist/main.py#L106)
So this code assumes that we have same number of data points in each batch, which actually not the case. And more accurate way should be multiply NNL_loss of each batch by its batch-size, and get the sum over whole data set, then divide it by the data set size.
2. The parameter `epoch` in [`test()` function](https://github.com/pytorch/examples/blob/master/mnist/main.py#L92) is useless. So we may as well remove it.

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -99,12 +99,11 @@ def test():
         data, target = Variable(data, volatile=True), Variable(target)
         current_batch_size = data.data.size()[0]
         output = model(data)
-        # F.nll_loss is averaged among each batch
-        test_loss += F.nll_loss(output, target).data[0] * current_batch_size
+        test_loss += F.nll_loss(output, target).data[0] * current_batch_size # sum up batch loss
         pred = output.data.max(1)[1] # get the index of the max log-probability
         correct += pred.eq(target.data).cpu().sum()
 
-    test_loss = test_loss # sum of loss function
+    test_loss = test_loss # sum of loss function over all data points
     test_loss /= len(test_loader.dataset)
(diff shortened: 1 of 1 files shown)


mnist/main.py
```diff
@@ -89,26 +89,27 @@ def train(epoch):
...
+        current_batch_size = data.data.size()[0]
         output = model(data)
-        test_loss += F.nll_loss(output, target).data[0]
+        test_loss += F.nll_loss(output, target).data[0] * current_batch_size # sum up batch loss
         pred = output.data.max(1)[1] # get the index of the max log-probability
         correct += pred.eq(target.data).cpu().sum()
 
-    test_loss = test_loss
-    test_loss /= len(test_loader) # loss function already averages over batch size
+    test_loss = test_loss # sum of loss function over all data points
```
apaszke: This line can be removed

mnist/main.py
```diff
@@ -89,26 +89,27 @@ def train(epoch):
...
     test_loss = 0
     correct = 0
     for data, target in test_loader:
         if args.cuda:
             data, target = data.cuda(), target.cuda()
         data, target = Variable(data, volatile=True), Variable(target)
+        current_batch_size = data.data.size()[0]
         output = model(data)
-        test_loss += F.nll_loss(output, target).data[0]
+        test_loss += F.nll_loss(output, target).data[0] * current_batch_size # sum up batch loss
```
apaszke: you can just use `size_average=False`

chao1224: Thanks, this makes codes more beautiful.

PR #176, opened 2017-07-06
added a function makedirs() to snli example which works both for python 2 and 3

andreh7:
since there is #145, here is another pull request to fix the problem that the `exist_ok` parameter of `os.makedirs()` does not exist in python < 3.2 (in particular it does not exist in python 2.7)  .


soumith:
thanks!

PR #179, opened 2017-07-07
fix bugs in generalization error calculation

chao1224:
Similar problem with #175 
Fix same issue in `mnist_hogwild.py`


soumith:
thanks!

PR #186, opened 2017-07-21
Fix test data in time_sequence_prediction

PeterChe1990:
Previously the training data input[:3] is incorrectly taken as the test data. Change the input for prediction to data[:3].


soumith:
thank you!

PR #189, opened 2017-07-23
Use nn.init from core to init SR example

### LINKED ISSUE #161: [super_resolution]
def _get_orthogonal_init_weights(weights):
    fan_out = weights.size(0)
    fan_in = weights.size(1) * weights.size(2) * weights.size(3)
    u, _, v = svd(normal(0.0, 1.0, (fan_out, fan_in)), full_matrices=False)
    if u.shape == (fan_out, fan_in):
        return torch.Tensor(u.reshape(weights.size()))
    else:
        return torch.Tensor(v.reshape(weights.size()))

Why do the above operation？

Kaixhin:
Makes it clearer what is going on in the init as well.
Closes #161

### DIFF
--- a/super_resolution/model.py
+++ b/super_resolution/model.py
@@ -1,20 +1,9 @@
 import torch
 import torch.nn as nn
+import torch.nn.init as init
 from numpy.random import normal
 from numpy.linalg import svd
 from math import sqrt
-
-
-def _get_orthogonal_init_weights(weights):
-    fan_out = weights.size(0)
-    fan_in = weights.size(1) * weights.size(2) * weights.size(3)
-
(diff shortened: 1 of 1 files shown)


soumith:
awesomesauce!

lanpa:
@Kaixhin  Does this modification work with standard pytorch release download from conda?

```
  File "~/examples/super_resolution/model.py", line 27, in _initialize_weights
    init.orthogonal(self.conv1.weight, init.gain('relu'))
AttributeError: module 'torch.nn.init' has no attribute 'gain'
```

Kaixhin:
@lanpa Sorry, that was a typo that should be fixed by https://github.com/pytorch/examples/pull/192/.

PR #191, opened 2017-07-31
End-to-end Memory networks example for pytorch

uditsaxena:
This is an implementation of the end to end memory networks in pytorch from [End-To-End Memory Networks](http://arxiv.org/abs/1503.08895) using the Facebook [bAbI](http://arxiv.org/abs/1502.05698) dataset.

### DIFF
--- a/memory_network_n2n/README.md
+++ b/memory_network_n2n/README.md
@@ -3,7 +3,7 @@ This repository contains a pytorch implementation of an End to End Memory Networ
 
 The model uses the method described in [End-To-End Memory Networks](http://arxiv.org/abs/1503.08895).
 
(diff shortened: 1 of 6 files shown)


frankxu2004:
I'm wondering what is the current status of this example?

uditsaxena:
The PR is still open for this to be reviewed and merged.

jekbradbury:
Can you explain why the temporal encoding parameters include batch size as a dimension? It's pretty strange in deep learning to have explicit batch-dependence like that other than through batchnorm.

bkj:
FWIW -- this throws a bunch of warnings w/ `pytorch==0.4.0` and `python==3.6.5`

PR #197, opened 2017-08-10
test_loader should use test_batch_size.

hiroyachiba:


alykhantejani:
This has actually been fixed in #213 so I think this can be closed cc @soumith

PR #200, opened 2017-08-14
reinforcement_learning fix reward threshold

Pierre-Bartet:
Since the gym package has changed, maximum step number is 200 in the cartpole environment, so the mean reward was unable to reach 200, and both reinforce and actor_critics scripts were never ending.
This fix replace the hard coded 200 step by env.spec.reward_threshold.

In the future and to avoid such cases were an updated package broke things, maybe 
`pip freeze > requirements.txt`
would help by freezing the current (and therefore working) version of each package ?


soumith:
thanks for the fix @Pierre-Bartet !

PR #202, opened 2017-08-15
Allow more gpu/process configurations in distributed imagnet example

csarofeen:


csarofeen:
This may be okay for gloo across processes but will not work if we support nccl as backend.

PR #203, opened 2017-08-16
Add fp16 support to word_language_model and imagenet

csarofeen:


csarofeen:
Had a mistake which broke distributed. Fixed now.

imagenet/fp16util.py
```diff
@@ -0,0 +1,29 @@
...
+
+def set_grad(params, params_with_grad):
+
+    for param, param_w_grad in zip(params, params_with_grad):
+        if param.grad is None:
+            param.grad = torch.nn.Parameter(param.data.new().resize_(*param.data.size()))
+        param.grad.data.copy_(param_w_grad.grad.data)
+
+
+def replace_with_BN16():
```
apaszke: No, let's not add such hacks to examples

imagenet/main.py
```diff
@@ -200,16 +232,29 @@ def train(train_loader, model, criterion, optimizer, epoch):
...
+        loss = loss*args.loss_scale
         # compute gradient and do SGD step
-        optimizer.zero_grad()
-        loss.backward()
-        optimizer.step()
+
+        if args.fp16:
+            model.zero_grad()
+            loss.backward()
+            if args.loss_scale != 1:
```
apaszke: You could just do `(loss / args.loss_scale).backward()` and it would be much faster. BTW it's weird that it's in fact _inverse scaling_ of the loss

imagenet/fp16util.py
ngimel: 1) fp16 bn in core is currently broken
2) even if it were not, for convergence you need fp32 parameters and buffers.

imagenet/main.py
csarofeen: Loss scaling is to help maintain a well behaved range during back propagation, preventing gradients from getting truncated to 0. Running `(loss / args.loss_scale).backward()` would be the same as not having loss scaling.

jekbradbury:
The master copy/update strategy is essentially mandatory for FP16, so there should probably be an "official" way to do it that's compatible with builtin optimizers?

ngimel:
Imagenet example shows how to use built-in optimizer on a master copy of parameters. It is hacky, but I don't know how palatable it would be to hide all this in the built-in optimizers.

PR #208, opened 2017-08-25
Change some parameters in mnist

yueruchen:
These parameter can made the accuracy up to 99% which is 98% before.


msaroufim:
Updated parameters are now available in main

PR #209, opened 2017-08-27
network structure diagram for sine wave example

andreh7:
This is a proposal to add a .png file with the (folded) network structure implemented by time_sequence_prediction/train.py .

If another file format is more suitable, let me know -- I implemented this diagram using pyx, I can also put the graph generating python code here.

(graphviz does not support orthogonal edges with record nodes well hence the use of pyx)


msaroufim:
Hi @andreh7 I recognize this a very late review but do you mind adding your image to the appropriate README.md and showing me what it looks like?

andreh7:
Let's abandon it, this is too long ago...

PR #213, opened 2017-09-05
Change test DataLoader to use the test batch size

rdinse:
The batch size for the training set was erroneously used for instantiating the DataLoader class for the test set.


soumith:
thanks!

PR #220, opened 2017-09-22
vae: Fix `UserWarning`

r9y9:
Fixes:

```
UserWarning: Using a target size (torch.Size([16, 1, 28, 28])) that is
different to the input size (torch.Size([16, 784])) is deprecated.
Please ensure they have the same size.
```

on PyTorch v0.2.0.

Second commit removes unnecessary branch I found during debugging.


r9y9:
Found wrong help messages. Fixed.

soumith:
thank you! cleaner!

PR #222, opened 2017-09-25
h rather than c should be fed into the next layer of LSTM

dtolpin:
The time series prediction example is somewhat misleading. In a multilayer LSTM, h (the hidden state or output) rather than c should be fed into the next layer or output as the prediction. For the sine wave, the network learns to predict either way, but I believe the example should be fixed to show the correct use.


soumith:
thank you!

PR #223, opened 2017-09-26
[BUG] potential typo in default learning rate

flennerhag:
Training any baseline model with default settings cause exploding loss or no learning at all. 

The default is set at 20. Setting it to something like 1 or 2 allows the models to learn.

### DIFF
--- a/word_language_model/main.py
+++ b/word_language_model/main.py
@@ -19,7 +19,7 @@ parser.add_argument('--nhid', type=int, default=200,
                     help='number of hidden units per layer')
 parser.add_argument('--nlayers', type=int, default=2,
                     help='number of layers')
-parser.add_argument('--lr', type=float, default=20,
+parser.add_argument('--lr', type=float, default=2,
                     help='initial learning rate')
 parser.add_argument('--clip', type=float, default=0.25,
                     help='gradient clipping')


soumith:
the default setting is border-line fastest learning rate we can use, i believe.

flennerhag:
Ok, I pushed this just in case it was a typo.

I suppose the default has been tested so should work. Currently I only have access to my laptop (mac), which might be a fringe case. In this case, running ``RNN_RELU`` with defaults arguments produces ``nan`` losses immediately. Running ``RNN_TANH`` with defaults causes training to plateau at a loss of roughly 10 (> 1000 in ppl). The losses I'm referring to are the printouts so the actual loss used by ``torch`` in the backprop step might not suffer from overflow, but regardless on my machine the loss is unstable and doesn't improve as expected.

Being examples, it might be an idea to lower the learning rate a bit to ensure training is stable, but that's of course up to you. If you're happy with the current default, feel free to close this PR.

apaszke:
I'd add gradient clipping. Exploding gradients are common in RNNs, and would happen even with a smaller LR (unless you actually make it extremely small)

apaszke:
I think we'll leave the defaults as they are. We're aware they won't work for all kinds of models, but that's how it is with hyperparameters.

flennerhag:
Sorry forgot about this. Actually, I found that the problem has to do with calculating the ppl score for the batches early in the first epoch. If the loss starts out at or above ``700``, this will cause an ``OverflowError`` when calling ``math.exp`` that aborts the entire program.

Putting a ``try-except`` block around calculating the perplexity score (``math.exp(cur_loss)``) solved the issue. Let me know if you'd like me to fire up a PR for it.

PR #226, opened 2017-10-01
Balance VAE losses, add reconstruction + sampling

Kaixhin:
Apparently this [wasn't giving decent results](https://discuss.pytorch.org/t/pytorch-vae-variational-autoencoder-example-not-training-no-meaningful-model-returned/8095) so I used the same balancing factor as I did for my Autoencoders repo.

This also saves images of reconstructions on the test set as well as decoded sample ~ N(0, 1). Showing my results after 1 epoch below, looks better now.

Reconstructions:
reconstruction_1

Samples:
sample_1


soumith:
sweet! much thanks!

joekidd:
@Kaixhin could you explain me please, what was the reason of modifying the reparametrize method in the way you did - now you are simply using trained means and you loose all the stochastic part? Maybe I'm missing something here, but it doesn't look like VAE to me anymore.

Kaixhin:
@joekidd The only time the trained means are used without the sampling is during evaluation when you want to reconstruct an input image. The interpretation is that you're recovering the latent sample `z` with the highest probability under the posterior `q(z|x)`. You could still sample `z`'s if you wanted to.

joekidd:
@Kaixhin that's perfectly makes sense! Thanks!

PR #228, opened 2017-10-01
Fix VAE loss + improve reconstruction viz

### LINKED ISSUE #225: VAE loss
According to the expression in line 95, the KL-divergence term is calculated from
`0.5 * sum(1 + log(sigma^2) - mu^2 - sigma^2)`
but I think the code in line 96-97 represents
`0.5 * sum(1 + log(sigma^2) - mu^2 - sigma)`

This might not be essential because whether the last term is squared or not, the loss descending behavior stays unchanged.

Kaixhin:
- Fixes VAE loss/makes it easier to read. Closes #225.
- Moves generations into separate folder (neater).
- Plots ground truth against test reconstructions for direct comparison (see below for example after 1 epoch).

reconstruction_1

Note: Forgot how bad fully connected VAE can be in comparison to convolutional VAE (w/ or w/o batch norm). These kind of results are in line with the original work.


soumith:
awesomesauce, thanks!

PR #229, opened 2017-10-01
Remove unused math import

### LINKED ISSUE #204: Unused import of math in time_sequence_prediction example
The generate_sine_wave.py module imports math on the first line, but doesn't use it. This import should be removed.

Kaixhin:
Closes #204


soumith:
thanks!

PR #235, opened 2017-10-15
add linux kernel and sampling to word language model

waynemystir:


apaszke:
Thanks for the PR, but I think that's too much much. This repo is meant to contain simple examples that show people how to write certain classes of models, without having a complete collection of relevant datasets.

waynemystir:
No problem. I thought the Linux kernel might be simpler and a bit more concrete than the Penn data, particularly with the addition of sample.py to show results. But I understand your point.

PR #249, opened 2017-11-04
Update RL examples to use torch.distributions instead of reinforce

Kaixhin:
cc @hughperkins to check

### DIFF
--- a/reinforcement_learning/actor_critic.py
+++ b/reinforcement_learning/actor_critic.py
@@ -8,8 +8,8 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 import torch.optim as optim
-import torch.autograd as autograd
 from torch.autograd import Variable
+from torch.distributions import Multinomial
 
 
 parser = argparse.ArgumentParser(description='PyTorch actor-critic example')
@@ -29,7 +29,9 @@ env.seed(args.seed)
 torch.manual_seed(args.seed)
 
 
-SavedAction = namedtuple('SavedAction', ['action', 'value'])
+SavedAction = namedtuple('SavedAction', ['log_prob', 'value'])
+
+
 class Policy(nn.Module):
     def __init__(self):
         super(Policy, self).__init__()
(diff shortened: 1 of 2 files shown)


soumith:
will merge when new binaries are shipped.

reinforcement_learning/actor_critic.py
```diff
@@ -54,29 +56,28 @@ def forward(self, x):
 def select_action(state):
     state = torch.from_numpy(state).float().unsqueeze(0)
     probs, state_value = model(Variable(state))
-    action = probs.multinomial()
-    model.saved_actions.append(SavedAction(action, state_value))
+    m = Multinomial(probs)
+    action = m.sample()
+    model.saved_actions.append(SavedAction(m.log_prob(action), state_value))
```
hughperkins: Cool. Looks good. I guess we can unwrap `state_value`, since we wont backprop through it?

Kaixhin: We will backprop through it - the critic needs updating (via the value loss). We just don't want the critic to be accidentally updated by the policy loss, so the policy loss uses the unwrapped value.

reinforcement_learning/reinforce.py
```diff
@@ -75,9 +77,9 @@ def finish_episode():
 running_reward = 10
 for i_episode in count(1):
     state = env.reset()
-    for t in range(10000): # Don't infinite loop while learning
+    for t in range(10000):  # Don't infinite loop while learning
         action = select_action(state)
-        state, reward, done, _ = env.step(action[0,0])
+        state, reward, done, _ = env.step(action[0, 0])
```
zou3519: It looks like action is now a 1D tensor. Does action[0, 0] work?

zou3519:
@Kaixhin regarding `reinforce.py`:

I ran the model under pytorch v0.3.0 and it doesn't seem to be training anymore:
image
I don't know much about RL, do you know what could be causing this?

Kaixhin:
@zou3519 I'm tempted to say that this is just the instability of pure REINFORCE. I checked the old `reinforce.py` under 0.2 - works fine - and `actor_critic.py` under 0.3 - works fine, so I can replicate your findings. I've tried tuning some hyperparameters, network architecture, the random seed etc., and can improve or worsen stability, but still haven't found a setting that learns properly. Perhaps someone who isn't familiar with RL can spot a difference between `reinforce.py` and `actor_critic.py` that I'm missing, but I didn't spot any discrepancies (apart from learning rate, which is something that would be tuned separately anyway).

jekbradbury:
cc @ebetica, who spent a while tuning both of these examples early on

hughperkins:
So, I reckon there is a bug in the `master` version of reinforce too. I've run a ton of REINFORCE algos lately, and sure there's high variance, but I've never seen the reward suddenly flatline to zero, and stay at exactly 0, forever.

But the one time I ported my model to `master` reinforce, and left it running overnight, I came back in the morning, and it had been flatlined pretty much the last 7 hours or so :P died at around 1am or so.

Hence, I left `master` alone for now, and am keeping on v0.2, so I don't have to be the one to diagnose/fix this bug :)  (sometimes I am energetic and fix stuff, but sometimes I try to stay focused on my actual intended end task :) ). But I do think there's something not quite right; and that correlates wth zou3519 noticing something odd plausibly.

colesbury:
There was a bug in the `reinforce.py` example. It referred to `reward` instead of `r`.

Kaixhin:
@colesbury well spotted, and thanks for tidying this up more generally too! Out of curiosity, are there any reasons one might want to accumulate losses in an array to then sum over rather than accumulating them on the fly using `+=`?

Issue #243, opened 2017-10-25
Time sequence prediction still broken


plang85:
Backed out #222 and it seems to work again.

plang85:
Just saw that this is still being duscussed in a previously closed issue #215

plang85:
In it's current state the example is broken, so I think this deserves to be an open issue, pls correct me if I'm wrong

plang85:
nevermind me, all works as advertised if installed using conda pm

vgoklani:
bump

barnabytprowe:
Thanks for linking this issue @plang85 .

I had been having trouble with this example and had put some results on #215.  I had to reduce the learning rate `lr` significantly to achieve a "good" solution.

I ran three cases spanning the decade from `lr` = 0.01 up to `lr` = 0.10.

These are below (gzipped avi and stdout output from the code):
lr = 0.01:
[lr_0.01_epochs_300_predicts.avi.gz](https://github.com/pytorch/examples/files/1435747/lr_0.01_epochs_300_predicts.avi.gz)
[lr_0.01_epochs_300_predicts.stdout.txt.gz](https://github.com/pytorch/examples/files/1435748/lr_0.01_epochs_300_predicts.stdout.txt.gz)

lr = 0.03:
[lr_0.03_epochs_300_predicts.avi.gz](https://github.com/pytorch/examples/files/1435749/lr_0.03_epochs_300_predicts.avi.gz)
[lr_0.03_epochs_300_predicts.stdout.txt.gz](https://github.com/pytorch/examples/files/1435750/lr_0.03_epochs_300_predicts.stdout.txt.gz)

lr = 0.10:
[lr_0.10_epochs_300_predicts.avi.gz](https://github.com/pytorch/examples/files/1435751/lr_0.10_epochs_300_predicts.avi.gz)
[lr_0.10_epochs_300_predicts.stdout.txt.gz](https://github.com/pytorch/examples/files/1435752/lr_0.10_epochs_300_predicts.stdout.txt.gz)

The videos are 10 steps / epochs per second, 300 epochs / 30 seconds total. If anyone would rather have the videos in a .zip format let me know.

As you can see, only the `lr` = 0.01 case converges to a solution which repeats in the way illustrated in the example. Both others find a different stable minimum in the cost function, in the `lr` = 0.10 case with a lower final cost than the "better" case that repeats stably.

I think
- when training, better fits to the training data can often be models that don't repeat well
[...]

PR #250, opened 2017-11-09
Add linear layer to time series prediction

t-vi:
As is the final network output is modulated with a tanh nonlinearity.
This is undesirable. As a simple / realistic fix we add a final
linear layer.

This addresses #243.

### DIFF
--- a/time_sequence_prediction/train.py
+++ b/time_sequence_prediction/train.py
@@ -12,23 +12,26 @@ class Sequence(nn.Module):
     def __init__(self):
         super(Sequence, self).__init__()
         self.lstm1 = nn.LSTMCell(1, 51)
-        self.lstm2 = nn.LSTMCell(51, 1)
+        self.lstm2 = nn.LSTMCell(51, 51)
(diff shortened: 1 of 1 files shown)


soumith:
thanks a lot for investigating and fixing it Thomas!

barnabytprowe:
Nice!

barnabytprowe:
A quick question @t-vi , as someone interested to learn - was the tanh nonlinearity in the output a bad idea in this specific example, or is it one _in general_?

t-vi:
@barnabytprowe, it really depends on the nature of what you are learning. Here, the tanh would have required the input to the tanh to get extremely large/small to reproduce values at +/- 1 or very close. This - that you need extremely large values - is a bad idea in general.
On the other hand, if you know that values close to +/-1 are extremely improbable, you might go with tanh at the end.
(we might move to the [discussion forums](https://discuss.pytorch.org/), I'm tom there)

barnabytprowe:
A belated thank you for the explanation @t-vi / Tom - makes total sense

PR #260, opened 2017-11-27
Swap PTB for Wikitext-2 (which is open access)

adamlerer:
We can't have PTB in Pytorch; swap for Wikitext-2.

@EdouardGrave


soumith:
cc: @Smerity

PR #266, opened 2017-12-08
fix reinforcement_learning example

blin00:
`torch.distributions.Multinomial` doesn't exist in v0.3.0, but `torch.distributions.Categorical` does


soumith:
thank you!

PR #268, opened 2017-12-14
Inception_v3 takes a different input size.

CHLLHC:
Inception_v3 takes an input of 299\*299\*3 instead of 224\*224\*3. This needs to be changed, otherwise, it will complain "...output size is too small..." when running the inception_v3 model.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -116,10 +116,14 @@ def main():
     normalize = transforms.Normalize(mean=[0.485, 0.456, 0.406],
                                      std=[0.229, 0.224, 0.225])
 
+    model_input_size = 224
+    if args.arch == 'inception_v3':
+        model_input_size = 299
+
     train_dataset = datasets.ImageFolder(
         traindir,
         transforms.Compose([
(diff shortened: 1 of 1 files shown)


happsky:
Inception_v3 still does not work after changing input size, any solutions?

CHLLHC:
It worked on my side. What is your error code? Maybe I have seen it.

rolloff:
This is implicitly setting the default resize length to 244, when before it was 256.

imagenet/main.py
```diff
@@ -136,8 +140,8 @@ def main():
 
     val_loader = torch.utils.data.DataLoader(
         datasets.ImageFolder(valdir, transforms.Compose([
-            transforms.Resize(256),
-            transforms.CenterCrop(224),
+            transforms.Resize(model_input_size),
```
kklt007: This line does not need to be modified.

soumith: yea this has to be 256

msaroufim:
Closing since no response from OP

PR #270, opened 2017-12-18
Typo/unused parameter in snli/train.py

pminervini:
There's a small typo in `snli/train.py`; incidentally, that parameter is not used anywhere, so I just replaced it with `_`.


msaroufim:
Now fixed on main

PR #284, opened 2018-01-13
Update main.py

zzy123abc:
parser.add_argument('--dist-rank', default=0, type=int,
                    help='rank of distributed processes')

if args.distributed:
        dist.init_process_group(backend=args.dist_backend, init_method=args.dist_url,world_size=args.world_size,rank=args.dist_rank)


zzy123abc:
init_process_group() need a rank now.

Issue #234, opened 2017-10-14
VAE Loss: The weight of BCE vs. KL


huanghoujing:
The BCE loss is averaged over the **batch dimension**, but KL is averaged over both **batch** and **image pixel dimension**. I hope to know the reason for this. 

I tried both 
- averaging the BCE loss over both dimensions 
- averaging the KL loss just over batch dimension

But the reconstructed images and generated images in test time are rather faint and blurry (just like digit 8), which is far worse than the setting here.

If the setting here is just to balance between two loss, why do you use the number of image pixels (which is intuitively supposed to have some mathematical meaning) for scaling. Why not use some number like 10, 100 or 1000?

Thanks for help.

huixutech:
According to [torch.nn.BCELoss](http://pytorch.org/docs/master/nn.html?highlight=binary_cross_entropy#torch.nn.BCELoss) doc,  and [its cpu implementation source code ](https://github.com/pytorch/pytorch/blob/a64560c22eb3882bc6b9ab15d12e1d832f3735fc/torch/legacy/nn/BCECriterion.py#L20-L54), if `size_average=True`, BCE loss is averaged over both batch and image pixel dimension. Because when computing `- log(input) * target - log(1 - input) * (1 - target)`, both `input` and `target` are resized into one dimension via `view(-1)`, and the outcome is divided by the one dimension size, which is `batch_size * image pixel dimension`.

huanghoujing:
I am still confused about how we should set the weight between two losses. 

Anyway, thank you very much!

Issue #290, opened 2018-01-22
Incorrect normalization in VAE example loss function


richardzhu:
In the `loss_function` part of the VAE example, I noticed that

`KLD = -0.5 * torch.sum(1 + logvar - mu.pow(2) - logvar.exp())
    # Normalise by same number of elements as in reconstruction
    KLD /= args.batch_size * 784
`

But the dimensionality of the latent variables (logvar, mu) is 20, not 784 -- hence it should either be
`torch.sum` and normalize by `args.batch_size * 20` or just straight-up `torch.mean`, otherwise the BCE and KLD losses are not properly scaled against each other. Changing the normalization from 784 to 20 increases the test error at the end of training, but this is due to a lower normalization increasing the scale of the KLD.

richardzhu:
This is related to issue #234!

soumith:
closed via https://github.com/pytorch/examples/pull/296

PR #296, opened 2018-02-02
Fix VAE losses (sum over everything)

Kaixhin:
Summation is consistent with notation in the paper. Results are still fine (test reconstructions and samples, noting that this VAE doesn't use conv layers and so results won't necessarily be as good):

reconstruction_10
sample_10

Closes https://github.com/pytorch/examples/issues/234 and https://github.com/pytorch/examples/issues/290.


soumith:
thanks Kai!

PR #300, opened 2018-02-08
Mnist

jindingars:


msaroufim:
I believe this is outdated at this point so closing

PR #301, opened 2018-02-08
Fix actor_critic example

colesbury:
```
 * smooth_l1_loss now requires shapes of arguments to match
 * once scalars are enabled, we must use torch.stack() instead of
   torch.cat() to turn list of scalars into a Tensor
```

### DIFF
--- a/reinforcement_learning/actor_critic.py
+++ b/reinforcement_learning/actor_critic.py
@@ -46,7 +46,7 @@ class Policy(nn.Module):
         x = F.relu(self.affine1(x))
         action_scores = self.action_head(x)
         state_values = self.value_head(x)
-        return F.softmax(action_scores, dim=1), state_values
+        return F.softmax(action_scores, dim=-1), state_values
 
 
 model = Policy()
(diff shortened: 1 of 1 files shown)


reinforcement_learning/actor_critic.py
```diff
@@ -56,6 +56,7 @@ def forward(self, x):
 def select_action(state):
     state = torch.from_numpy(state).float().unsqueeze(0)
     probs, state_value = model(Variable(state))
+    state_value = state_value.squeeze(0)
```
gchanan: I think this will work if you just get rid of the unsqueeze above.

colesbury: Done

LUKELIEM:
I tried to adapt your code to run Pong and it runs into severe memory issues. With cuda, I encountered cuda runtime error: out of memory. Without cuda, the Linux PC run out of RAM (32G).

In a Pong Game, where the agent plays up to 1500 steps with the game AI, the simple 2-layer NN with softmax seems to be replicated 1500 times in the computational graph. Is there a more memory efficient implementation of both REINFORCE and ACTOR-CRITIC?

PR #307, opened 2018-02-22
Fixes for Tensor and Variable merge

colesbury:
Fixes some warnings when indexing a 0-d array and potential overflow since Variable `sum()` returns the same type as the source tensor


yf225:
Would this break PRs that comes before https://github.com/pytorch/pytorch/pull/5225? One thing we can do is to change the MNIST links in https://github.com/pytorch/pytorch/blob/master/.jenkins/perf_test/test_gpu_speed_mnist.sh and https://github.com/pytorch/pytorch/blob/master/.jenkins/perf_test/test_cpu_speed_mnist.sh to point to colesbury:perftests in the PyTorch PR, and then after it is merged we can merge this PR to master.

colesbury:
No, it should be backwards compatible (even with 0.3)

PR #310, opened 2018-02-24
Adjust the loss according to batch size

samuela:
This way you don't get weird jumps when you plot the ELBO.


samuela:
Seems to be addressed on master now.

PR #313, opened 2018-03-04
bug fix for MNIST conv net example

wjmaddox:
Correction to small bug in the calculation of the testing accuracy. Manually inspecting the the pred and target tensors will show the bug.

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -91,8 +91,8 @@ def train(epoch):
 
 def test():
     model.eval()
-    test_loss = 0
-    correct = 0
+    test_loss = 0.0
+    misclass = 0.0
     for data, target in test_loader:
         if args.cuda:
             data, target = data.cuda(), target.cuda()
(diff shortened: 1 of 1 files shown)


apaszke:
What was the bug? It looks like the patch just changes accuracy to error rate, but they are equivalent

wjmaddox:
The bug is in the equality counter. If you print out a concatenated version of the current pred and target variables, you'll see that they align more often than correct says that they do.

I changed accuracy to error rate, which are equivalent.

apaszke:
Was this caused by broadcasting? I really can't spot the error in the original code. I think accuracy is a more standard metric when running scripts on MNIST, so let's stick to that

wjmaddox:
Not quite, it looks like it was caused by the fact that .eq returns a ByteTensor and not a Float or LongTensor. Then, summing over this ByteTensor will give the answer mod 256.

A simpler correction might just be changing line 103 to 
`correct += pred.eq(target.data.view_as(pred)).long().cpu().sum()`

Sorry about changing to accuracy instead of error.

soumith:
thanks for pointing the bug. it looks like we just have to add `.long()`. I've fixed it in https://github.com/pytorch/examples/commit/c66593f1699ece14a4a2f4d314f1afb03c6793d9

PR #320, opened 2018-03-23
Minor style update: 'input' is a built-in, shouldn't be used a variable.

alanderex:
Minor style update: 'input' is a built-in, shouldn't be used a variable.


msaroufim:
I think it's fine `inpt` doesn't feel like the right name either

PR #321, opened 2018-03-26
Fix InstanceNorm

smessmer:
The example is broken with pytorch master. This applies the fix from here: https://github.com/pytorch/pytorch/issues/5602


SsnL:
will be fixed in #331

msaroufim:
Duplicate of #331

PR #325, opened 2018-04-05
Add --max-batches-per-epoch option to MNIST example

yf225:
The whole MNIST dataset takes ~90s to train on the new bare metal CPU perf machines. Adding `--max-batches-per-epoch` so that we can have less batches per epoch and get the perf result faster.

This will only change the code in the `perftests` branch.

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -6,6 +6,7 @@ import torch.nn.functional as F
 import torch.optim as optim
 from torchvision import datasets, transforms
 from torch.autograd import Variable
+import sys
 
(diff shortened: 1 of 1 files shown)


apaszke:
Can we keep the perf testing code a for of the examples? Those files are meant as demos/starter code for projects and this flag isn't useful anywhere outside this single use case

yf225:
Sure that makes sense - I will probably move things over to https://github.com/pytorch/benchmark

apaszke:
Thank you!

PR #328, opened 2018-04-10
Make the MNIST be compatible with the PyTorch v0.5

Sanghyun-Hong:


msaroufim:
mnist is now working with the most recent version of pytorch

PR #333, opened 2018-04-20
Update the script for word language model

houseroad:
Let it work with master.


houseroad:
I see we already have the change pr, nice!

PR #340, opened 2018-04-25
Update main.py

laoreja:
setting input to cuda, fix minor bugs.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -186,7 +186,8 @@ def train(train_loader, model, criterion, optimizer, epoch):
         # measure data loading time
         data_time.update(time.time() - end)
 
-        target = target.cuda(True)
+        input = input.cuda()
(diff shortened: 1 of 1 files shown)


soumith:
we dont need to move input to cuda, DataParallel will do it better.
If input is on CPU, DataParallel will scatter parts of input directly to respective GPUs.
i.e.

[batch split 1] cpu -> cuda0
[batch split 2] cpu -> cuda1

laoreja:
Thank you very much!
May I ask why part of these two models are not in DataParallel:
if args.arch.startswith('alexnet') or args.arch.startswith('vgg'):
            model.features = torch.nn.DataParallel(model.features)
            model.cuda()

soumith:
for those models, they have very large fully-connected layers, which are not good for DataParallel.

PR #346, opened 2018-05-02
Add Continuous bag of words example

FraLotito:
This PR adds an implementation of the CBOW model, it's a nice way to start learning PyTorch


avinashsai:
@FraLotito 
Hi, As far as I know the hidden layer in Word2Vec models doesn't contain any activation function. Hidden layer weights actually represents the Word Vectors trained.Hence, no non-linearity function is added.

PR #353, opened 2018-05-12
Convert to proper script with main(), fix #352

### LINKED ISSUE #352: Multiprocessing runtime error freeze_support() in Windows 64 bit
copied from original issue from below , but looks like it has to be fixed in this examples:

https://github.com/pytorch/pytorch/issues/7485

sytelus:
This is simple restructure of code without any logic changes so that now it has proper main() function.

This is required because it otherwise does not run on Windows. This PR fixes issue: https://github.com/pytorch/examples/issues/352. More info on fix is [here](https://pytorch.org/docs/stable/notes/windows.html#multiprocessing-error-without-if-clause-protection).

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -24,39 +24,38 @@ class Net(nn.Module):
         x = self.fc2(x)
         return F.log_softmax(x, dim=1)
 
+def train(args, model, device, train_loader, optimizer, epoch):
+    model.train()
+    for batch_idx, (data, target) in enumerate(train_loader):
+        data, target = data.to(device), target.to(device)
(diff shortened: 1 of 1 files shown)


peterjc123:
Why do you wrap all the functions into `main`?  I don't think this is a good idea.

sytelus:
@peterjc123, @apaszke  - The train() and test() function shared the variables defined in main in original version so to keep changes to minimum I put them as sub functions. Anyway, they are now moved out of main() and accept variables they need from main() as params.

soumith:
thanks Shital!

PR #354, opened 2018-05-14
fix 0.4 version compliance

chroneus:
detach Variable before call numpy()


chroneus:
Fix runtime error

soumith:
thank you!

PR #364, opened 2018-06-01
[imagenet example]use ```lr_scheduler``` instead of ```adjust_learning_rate```

jxgu1016:
PyTorch already have this function ```lr_scheduler``` to modify the learning rate,  so why we still use user-customized function ```adjust_learning_rate```?


soumith:
this is done for coherence of example, i.e. to not introduce all concepts at once.

PR #381, opened 2018-07-06
Add support for deterministic behavior in ImageNet benchmark

jma127:
This was tested by running the script without the two extra flags (to verify existing functionality), then running the script with the two extra flags twice (to verify deterministic behavior).


jma127:
cc @SsnL and @soumith for review

imagenet/main.py
```diff
@@ -63,6 +68,22 @@ def main():
     global args, best_prec1
     args = parser.parse_args()
 
+    if args.seed is not None:
+        random.seed(args.seed)
+        torch.manual_seed(args.seed)
+        torch.cuda.manual_seed_all(args.seed)
+        cudnn.deterministic = True
+        print('WARNING: you have chosen to seed training.')
```
SsnL: nit: use warnings.warn instead of print

imagenet/main.py
```diff
@@ -63,6 +69,21 @@ def main():
     global args, best_prec1
     args = parser.parse_args()
 
+    if args.seed is not None:
+        random.seed(args.seed)
+        torch.manual_seed(args.seed)
```
soumith: `torch.manual_seed(args.seed)` will also seed CUDA, so you dont have to add `torch.cuda.manual_seed_all(args.seed)`

jma127:
Resolved the above comments.

jma127:
Thanks for reviewing :) . I don't seem to have write access to this repo, so could you merge?

SsnL:
Me neither :) Soumith's traveling, but he should be able to merge when he's near laptop.

PR #394, opened 2018-07-26
Fix snli device to work with newest torchtext

SsnL:


fcalvet:
Just to anyone seeing the error message 

> TypeError: nll_loss() got an unexpected keyword argument 'reduction'

This is due to the reduction keyword not being supported on 0.4.0 (see https://pytorch.org/docs/0.4.0/nn.html#nll-loss)
So updating to 0.4.1 should solve the error (even if it's not written in the release notes, it's in the docs)

PR #398, opened 2018-08-06
Some little change on  full-connect layer

an1006634493:
Use a new arg to make it more clear


surgan12:
I dont think this was needd

PR #401, opened 2018-08-15
new example of reinforcement learning with MountainCar enviroment

zergey:


msaroufim:
Hi @zergey could you please also add a test here https://github.com/pytorch/examples/blob/main/run_python_examples.sh

PR #402, opened 2018-08-19
super_resolution: add freeze_safe support, python lint issues, checkpoint_every_n option

jefflomax:
This is a simple update to super_resolution so that:

It implements standard "freeze_support" so multithreading works on both Windows and Linux
Removes Python-Lint warnings by passing references to variables
Adds the option to store a checkpoint every N epochs, defaulted to 1
Without these changes, the sample does not run on Windows.

### DIFF
--- a/super_resolution/main.py
+++ b/super_resolution/main.py
@@ -9,41 +9,8 @@ from torch.utils.data import DataLoader
 from model import Net
 from data import get_training_set, get_test_set
 
-# Training settings
-parser = argparse.ArgumentParser(description='PyTorch Super Res Example')
-parser.add_argument('--upscale_factor', type=int, required=True, help="super resolution upscale factor")
-parser.add_argument('--batchSize', type=int, default=64, help='training batch size')
-parser.add_argument('--testBatchSize', type=int, default=10, help='testing batch size')
-parser.add_argument('--nEpochs', type=int, default=2, help='number of epochs to train for')
-parser.add_argument('--lr', type=float, default=0.01, help='Learning Rate. Default=0.01')
-parser.add_argument('--cuda', action='store_true', help='use cuda?')
-parser.add_argument('--threads', type=int, default=4, help='number of threads for data loader to use')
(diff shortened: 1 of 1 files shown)


sammiie:
To address the "super_resolution: add freeze_safe support, python lint issues, checkpoint_every_n option" issue #402, you will need to make code changes and enhancements to your super_resolution project. Below, I'll provide code snippets and explanations for each part of the issue:

1:Add Freeze_Safe Support:
To add freeze_safe support, you can modify your training loop to freeze certain layers of your neural network during training. This is useful for transfer learning or fine-tuning. Here's a simplified example:

import torch
import torch.optim as optim

# Assuming 'model' is your neural network
optimizer = optim.Adam(model.parameters(), lr=learning_rate)

if freeze_safe:
    # Freeze specific layers (e.g., the first three layers)
    for param in model.parameters():
        param.requires_grad = False
    for param in model.fc1.parameters():
        param.requires_grad = True

Modify this code according to your specific neural network architecture and which layers you want to freeze.

2:Python Lint Issues:
Python linting issues typically refer to code style and formatting violations. You can use a tool like flake8, pylint, or black to automatically fix or identify code style issues. Here's how you can use black to format your code:

Install black if you haven't already:
pip install black
Format your code using black:
black your_code.py
This will automatically format your code to comply with the PEP 8 style guide

3:Add checkpoint_every_n Option:
To add the checkpoint_every_n option, you can modify your training loop to save checkpoints (model weights) at specific intervals during training. Here's an example using PyTorch:

for epoch in range(num_epochs):
    for batch in data_loader:
        # Training steps here

        if iteration % checkpoint_every_n == 0:
            # Save a checkpoint
            torch.save(model.state_dict(), f'checkpoint_epoch_{epoch}_iter_{iteration}.pt')

This code saves a checkpoint of the model every checkpoint_every_n iterations. You can adjust the frequency according to your needs.

Remember to incorporate these code changes into your existing project codebase and adapt them to your specific use case and neural network architecture.

PR #416, opened 2018-09-22
Change incorrect comment on accuracy metric

xamat:
The way accuracy is implemented is computing recall@k not precision@k as stated. Note that this change is only arguing for consistency between code and comment, but is not making an assessment on the correctness of computing recall vs. precision.


soumith:
fixed the wording as "accuracy", instead via https://github.com/pytorch/examples/commit/8473762c8824e3d546d050f7dfc13e40edf3a416

PR #419, opened 2018-10-05
Fix bug in reparameterize() in VAE example

dpkingma:
The reparameterize() function should not be deterministic for validation/test mode. It should be the same as in training mode, otherwise what is computed is not the ELBO (i.e. the correct loss) but something else. This change will slightly change the test set error.

### DIFF
--- a/vae/main.py
+++ b/vae/main.py
@@ -51,12 +51,9 @@ class VAE(nn.Module):
         return self.fc21(h1), self.fc22(h1)
 
     def reparameterize(self, mu, logvar):
-        if self.training:
-            std = torch.exp(0.5*logvar)
-            eps = torch.randn_like(std)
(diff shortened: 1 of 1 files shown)


soumith:
thanks Durk!

ggbioing:
I guess that this is due to the Jensen's inequality E_z[log(f(z))] < log(f(E[z])).
If in the testing you do not sample the log-likelihood term of the ELBO you end up with an higher ELBO in the testing with respect to the training.
Is the reasoning correct?

dpkingma:
Well, almost. Jensen's inequality states that E_z[g(z)] <= g(E[z]) if g()
is concave; so if g = log f() is concave than you would be correct, but we
don't know.

So in principle the bug could have resulted in either higher or lower
validation loss.

PR #420, opened 2018-10-09
Update all prec occurrences to acc in imagenet tutorial

vipinpillai:
This pull request is to update all the occurrences of precision to accuracy in imagenet/main.py to be consistent with the accuracy definition.


soumith:
thanks!

PR #422, opened 2018-10-12
Character level RNN example

pskrunner14:
PyTorch example that generates novel names from scratch and can be a useful resource for learning how to easily handle sequential data with the framework. It contains a character Level RNN (RNN, LSTM and GRU) language model implemented in PyTorch.


apaszke:
We already have the word level language modelling example, which is pretty much isomorphic to this one, except it works on words instead of on characters. I'm not sure if we really want to merge this.

pskrunner14:
@apaszke thanks for the feedback. Closing this.

PR #423, opened 2018-10-13
Use torchtext 0.3.1

nzw0301:
The default value of `repeat` of `iterator` class is `False` in [the current stable version of torchtext](https://github.com/pytorch/text/releases/tag/0.3.1).
So [this line](https://github.com/pytorch/examples/blob/master/snli/train.py#L61) is not necessary now.


soumith:
thanks!

PR #426, opened 2018-10-28
modified the hogwild example

shagunsodhani:
Modified the hogwild example to perform testing outside the different training processes based on discussion [here]](https://discuss.pytorch.org/t/hogwild-training/27614)


soumith:
thanks @shagunsodhani !

PR #432, opened 2018-11-01
Fix redundant unk token for labels, making SNLI 3 class classification again

sksq96:
- Including unk in `answers` Field, made this a 4 class classification problem instead of 3. Set unk token to `None`.
- Always run `model.to(device)`.

### DIFF
--- a/snli/train.py
+++ b/snli/train.py
@@ -18,7 +18,7 @@ torch.cuda.set_device(args.gpu)
 device = torch.device('cuda:{}'.format(args.gpu))
 
 inputs = data.Field(lower=args.lower, tokenize='spacy')
-answers = data.Field(sequential=False)
+answers = data.Field(sequential=False, unk_token=None)
 
 train, dev, test = datasets.SNLI.splits(inputs, answers)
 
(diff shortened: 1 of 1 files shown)


snli/train.py
```diff
@@ -18,7 +18,7 @@
 device = torch.device('cuda:{}'.format(args.gpu))
 
 inputs = data.Field(lower=args.lower, tokenize='spacy')
-answers = data.Field(sequential=False)
+answers = data.Field(sequential=False, unk_token=None)
```
msaroufim, line 21: I'm not sure how I feel about this doesn't feel necessary

snli/train.py
```diff
@@ -50,7 +50,7 @@
     model = SNLIClassifier(config)
     if args.word_vectors:
         model.embed.weight.data.copy_(inputs.vocab.vectors)
-        model.to(device)
+model.to(device)
```
msaroufim, line 53: Makes sense

msaroufim (changes requested):
This doesn't feel super critical, would need to see some more tests on whether `model` scoping works and why the `unk_token` change was needed. So closing for now but lemme if I'm wrong and please reopen.

PR #441, opened 2018-11-13
[Distributed] Multiprocessing Distributed Training for ImageNet Example

teng-li:
Added multiprocessing mode for distributed training. This will use `torch.multiprocessing.spawn` for both single node and multi-node training.

Tested by verifying the following works:

(1) Single Node Single GPU
(2) Single Node Multi GPU
(3) multiprocessing single node Multi GPU distributed
(4) non-multiprocessing single node multi GPU distributed

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -73,10 +73,6 @@ parser.add_argument('--multiprocessing-distributed', action='store_true',
                          'N processes per node, which has N GPUs. This is the '
                          'fastest way to use PyTorch for either single node or '
                          'multi node data parallel training')
-parser.add_argument('--benchmark', action='store_true',
-                    help='Benchmark mode, will use synthetic image data')
-parser.add_argument('--benchmark-iter', default=100, type=int,
-                    help='Number of iterations to benchmark')
 
 best_acc1 = 0
 
@@ -84,10 +80,6 @@ best_acc1 = 0
 def main():
     args = parser.parse_args()
 
-    if args.benchmark:
-        if args.benchmark_iter < 20:
-            raise RuntimeError("Requires a minimum of 20 iterations to "
-                               "benchmark")
     if args.seed is not None:
         random.seed(args.seed)
         torch.manual_seed(args.seed)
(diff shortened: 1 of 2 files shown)


teng-li:
@soumith all comments addressed

imagenet/main.py
```diff
@@ -83,12 +94,49 @@ def main():
...
 
-    args.distributed = args.world_size > 1
+    args.distributed = args.world_size > 1 or args.multiprocessing_distributed
+
+    ngpus_per_node = torch.cuda.device_count()
+    if args.multiprocessing_distributed:
+        # python version check
+        if not ((sys.version_info[0] == 3 and sys.version_info[1] >= 4) or
+                sys.version_info[0] > 3):
+            raise RuntimeError("Requires python 3.4 or higher to use "
```
soumith: can you move this into `mp.spawn`. There's a large error message here, a lot of logic, all of it not really driving a short example.

`mp.spawn` should be raising such an error anyways, and suggest to people the error that you wrote here.

imagenet/main.py
```diff
@@ -83,12 +94,49 @@ def main():
...
+
+    if args.gpu is not None:
+        print("Use GPU: {} for training".format(args.gpu))
+
+    if args.distributed:
+        if args.multiprocessing_distributed:
+            # For multiprocessing distributed training, rank needs to be the
+            # global rank among all the processes
+            args.rank = args.rank * ngpus_per_node + gpu
+        dist.init_process_group(backend=args.dist_backend,
```
soumith: nit: our line limits are 120 chars. why break the lines before we hit the limit.

imagenet/main.py
```diff
@@ -97,12 +145,32 @@ def main():
...
+            # ourselves based on the total number of GPUs we have
+            args.batch_size = int(args.batch_size / ngpus_per_node)
+            model = torch.nn.parallel.DistributedDataParallel(
+                model,
+                device_ids=[args.gpu],
+                output_device=None)
+        else:
+            model.cuda()
+            # DistributedDataParallel will divide and allocate batch_size to all
+            # available GPUs if device_ids are not not set
```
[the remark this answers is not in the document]
teng-li: Addressed

imagenet/main.py
teng-li: addressed

imagenet/main.py
teng-li: addressed

PR #444, opened 2018-11-16
Update to save trained model

cdiep:
Add code to save the trained model.  Also add arguments to allow users
to pass in data and result directory locations.


soumith:
closed instead via https://github.com/pytorch/examples/pull/469

PR #446, opened 2018-11-17
add --finetune option for resnet, vgg, and densenet architectures

dribnet:
The finetune option is used to freeze features layers on a pretrained network

refactored and updated code from pytorch/examples#58


soumith:
instead of making `main.py` a one-stop shop for everything, I feel like a `finetune.py` would be a much better idea here. What do you think?

msaroufim:
@dribnet I'm happy to merge this but it probably needs a test. Take a look at this for some inspiration https://github.com/pytorch/examples/blob/main/run_python_examples.sh

PR #449, opened 2018-11-21
Unified order of flag names in argparser

timothygebhard:
Unified order of flag names in argparser: short flag names ('-f') go before long flag names ('--foo').


soumith:
thanks!

PR #454, opened 2018-11-26
Fix broken args order

ntomita:
For multiple long options, python takes the first as attribute name by default. The latest commit swapped the order of options that causes the script an error (e.g., AttributeError: 'Namespace' object has no attribute 'weight_decay'). Added dest option to explicitly specify the valid attribute name.


soumith:
thank you!

PR #466, opened 2018-12-10
Fix a typo in the `actor_critic` example

zeakey:
Original:
```
parser.add_argument('--seed', type=int, default=543, metavar='N',
                    help='random seed (default: 1)')
```
Fixed:
```
parser.add_argument('--seed', type=int, default=543, metavar='N',
                    help='random seed (default: 543)')
```


soumith:
thank you @zeakey !

Issue #476, opened 2018-12-17
--resume fails after 1 epoch with Pytorch 1.0 release


mdflickner:

surgan12:
How about declaring best_acc in shared memory .
i.e: best_acc.share_memory()

PR #477, opened 2018-12-17
some minor fixes

surgan12:
fix to #476

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -5,7 +5,7 @@ import shutil
 import time
 import warnings
 import sys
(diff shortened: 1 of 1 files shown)


soumith:
@surgan12 i fail to see how this fixes https://github.com/pytorch/examples/issues/476
could you explain?

surgan12:
@soumith , I think the issue is with accessing  best_acc on multiple GPUs acrosss multiples processes , so I sort of made it accessible across all the process .
I am not sure if this is something working here .

PR #494, opened 2019-01-16
Use argmax() instead of max()[1] in mnist/main.py

seungwonpark:
Thanks for great examples!
For beginners, I thought using `.argmax()` will be better than using `.max()[1]`.


soumith:
thanks!

PR #495, opened 2019-01-20
enhancing readability

andrewnaguib:
Removing `sys` package as it wasn't used. Also, renaming all `input` instances to `predictor` as `input` is a builtin keyword.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -263,23 +263,23 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
     model.train()
 
     end = time.time()
-    for i, (predictor, target) in enumerate(train_loader):
+    for i, (images, target) in enumerate(train_loader):
         # measure data loading time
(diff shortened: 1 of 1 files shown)


soumith:
we are explicitly not planning to change `input`, as it's the best word that describes it. `predictor` doesn't even come close.

Removing the `sys` package sounds good.

andrewnaguib:
Maybe we can capitalize INPUT?

On Jan 21, 2019, at 01:05, Soumith Chintala <notifications@github.com<mailto:notifications@github.com>> wrote:


we are explicitly not planning to change input, as it's the best word that describes it. predictor doesn't even come close.

Removing the sys package sounds good.

soumith:
capitalizing input just looks weird. none of the other variables will be capitalized.

In the context of just this example, `images` might make sense.

soumith:
thanks!

PR #506, opened 2019-02-14
[C++ Frontend] own dataset example

nocotan:
This is an implementation example of own dataset using PyTorch C ++ Frontend.

The ```test_data``` is a pair of images and labels.

```bash
$ cat ../test_data/labels.txt
image1.jpg,0
image2.jpg,0
image3.jpg,1
```


soumith:
cc: @yf225

PR #509, opened 2019-02-14
Add dropout layer in reinforce.py

yanxurui:
1. Dropout layer can significantly improve the performance and the problem is solved within 1000 episodes. In contrast, the original code seems not converge even after 10 thousand episodes.
2. Shift to CartPole-v1 which is a more challenging task than v0.

Credit: https://medium.com/@ts1829/policy-gradient-reinforcement-learning-in-pytorch-df1383ea0baf

Thanks,
Xurui


soumith:
thanks!

PR #511, opened 2019-02-15
addition of cgan example

kmualim:
Added a cgan example to pytorch


soumith:
while the code is concise and good, we want pytorch/examples to not be overwhelming, which means only a few examples (and probably just 1 or 2 on a topic).
So I think you can keep it as a separate project, and add it to a list like https://github.com/ritchieng/the-incredible-pytorch

PR #513, opened 2019-02-19
Tidy var names for RL examples

Kaixhin:
In response to some confusion in Slack, I've revisited the RL examples. The code to track the `running_reward` was mixed with number of timesteps, which is true for this environment but not in general, so this now makes it explicitly tracking reward only (cc @ebetica). I've also replaced (where appropriate), `rewards` with `returns`, which is actually what is happening. Run the code, everything works fine.

### DIFF
--- a/reinforcement_learning/actor_critic.py
+++ b/reinforcement_learning/actor_critic.py
@@ -67,16 +67,16 @@ def finish_episode():
     saved_actions = model.saved_actions
     policy_losses = []
     value_losses = []
-    rewards = []
+    returns = []
     for r in model.rewards[::-1]:
         R = r + args.gamma * R
-        rewards.insert(0, R)
-    rewards = torch.tensor(rewards)
-    rewards = (rewards - rewards.mean()) / (rewards.std() + eps)
-    for (log_prob, value), r in zip(saved_actions, rewards):
-        reward = r - value.item()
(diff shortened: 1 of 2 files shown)


brando90:
can we avoid hard coded constants like: 
```
running_reward = 0.05 * ep_reward + (1 - 0.05) * running_reward
```

Kaixhin:
That's a good point @brando90 - you should go ahead and submit a PR to clear up anything like this.

brando90:
@Kaixhin  I think there are still mistakes. e.g.
```
            print('Episode {}\tLast reward: {:.2f}\tAverage reward: {:.2f}'.format(
                  i_episode, ep_reward, running_reward))
```
ep_reward is not an average but a sum of (nondiscoutned) rewards for the current episode, so saying average reward doesn't make sense to me.

brando90:
@Kaixhin why do you start ur running return with 10?

https://github.com/pytorch/examples/blob/bc8feda9cce9d23e39efbe414a0e14f0a2de7f55/reinforcement_learning/reinforce.py#L79

Kaixhin:
@brando90 That was simply there before so I didn't touch it. In this case I believe it's roughly the starting performance of the agent, but this should ideally be set to 0 or documented with a comment.

PR #515, opened 2019-02-22
Fixes bug when loading checkpoint of different GPU

cgnorthcutt:
Lines 245 and 246 will always FAIL when a checkpoint is saved while running this code on a specific GPU, then something happens stopping training, then training is resumed on a different GPU. The error occurs because best_acc1 and acc1 are on different GPUs but are compared. This is a least-number-of-lines bug fix that does not impact other use cases.


soumith:
thanks!

PR #528, opened 2019-03-18
Use lr_scheduler for ImageNet training

pmeier:
This implements the usage of a `torch.optim.lr_scheduler` to take care of the learning rate adjustments. It furthermore deletes the `checkpoint` after it is loaded to reduce the memory requirement if resuming from a previous run.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -166,12 +166,14 @@ def main_worker(gpu, ngpus_per_node, args):
         else:
             model = torch.nn.DataParallel(model).cuda()
 
-    # define loss function (criterion) and optimizer
+    # define loss function (criterion), optimizer, and learning rate scheduler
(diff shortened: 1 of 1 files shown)


soumith:
it was an explicit choice to not use LR schedulers when we wrote the imagenet example, as this particular user space often wants to tweak learning rates by hand.

Having a level of indirection on how to be able to tweak the LR by hand felt like it would miss the point for those sets of users who will go grappling to other avenues like the forums to find out how to do it.

pmeier:
@soumith Out of curiosity: what purpose does the `adjust_learning_rate` function serve, if we expect the user to adjust the learning rate on his own?

soumith:
@pmeier when I say "tweak the LR by hand", I mean "write the code by hand", rather than relying on an opaque class. By having `adjust_learning_rate` close the the rest of the code, in the same file, the user can see and modify immediately the function

PR #530, opened 2019-03-18
Add logging to ImageNet training

pmeier:
This adds logging to the console and a log file. For now the logger is configured to log infos to `stdout` and everything with a higher level to `stderr`. Furthermore each record shows the time and level in addition to the message.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -173,6 +173,9 @@ def main_worker(gpu, ngpus_per_node, args):
                                 momentum=args.momentum,
                                 weight_decay=args.weight_decay)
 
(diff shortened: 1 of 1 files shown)


soumith:
I think for a base example, adding specialized logging is a bit of an overkill. My main hesitation is to increase the line count of the file over something that's not very important for an example script.

It makes more sense in something like torchvision reference scripts though: https://github.com/pytorch/vision/pull/819

pmeier:
With the new reference scripts (which were comitted after I created this PR) you are indeed correct. I will open a new PR in `pytorch/vision`.

PR #541, opened 2019-04-10
[C++][DCGAN] Log print Error

bintonto:
The current log display is not displayed at the "kLogInterval" interval, but is displayed with the "kCheckpointEvery", now fix the problem


soumith:
thank you!

PR #545, opened 2019-04-15
reversed function

drozzy:
Replace reverse indexing with reversed python function. I think this is more legible for general audience.


drozzy:
Nevermind, the extended slicing syntax looks pretty good too:

https://docs.python.org/2.3/whatsnew/section-slices.html

PR #555, opened 2019-05-03
Apply Transformer model for the word language problem in pytorch/examples

zhangguanheng66:
Apply recently developed PyTorch Transformer model for the word language problem. Any reviews and suggestions are appreciated. Please don't land this PR until the transformer model is fully supported by PyTorch. @cpuhrsch @soumith

### DIFF
--- a/word_language_model/README.md
+++ b/word_language_model/README.md
@@ -4,11 +4,15 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) on a language modeli
 By default, the training script uses the Wikitext-2 dataset, provided.
 The trained model can then be used by the generate script to generate new text.
 
-```bash
-python main.py --cuda --epochs 6        # Train a LSTM on Wikitext-2 with CUDA
-python main.py --cuda --epochs 6 --tied # Train a tied LSTM on Wikitext-2 with CUDA
-python main.py --cuda --tied            # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs
-python generate.py                      # Generate samples from the trained LSTM model.
+The recent updates support a Transformer model, which is composed of encoder and decoder with multihead attention layers.
+
+```bash 
+python main.py --cuda --epochs 6           # Train a LSTM on Wikitext-2 with CUDA
+python main.py --cuda --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA
+python main.py --cuda --epochs 6 --model Transformer --lr 5   
+                                           # Train a Transformer model on Wikitext-2 with CUDA
+python main.py --cuda --tied               # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs
+python generate.py                         # Generate samples from the trained LSTM model.
 ```
 
 The model uses the `nn.RNN` module (and its sister modules `nn.GRU` and `nn.LSTM`)
@@ -21,24 +25,28 @@ The `main.py` script accepts the following arguments:
 
 ```bash
 optional arguments:
-  -h, --help         show this help message and exit
-  --data DATA        location of the data corpus
-  --model MODEL      type of recurrent net (RNN_TANH, RNN_RELU, LSTM, GRU)
-  --emsize EMSIZE    size of word embeddings
-  --nhid NHID        number of hidden units per layer
-  --nlayers NLAYERS  number of layers
-  --lr LR            initial learning rate
-  --clip CLIP        gradient clipping
-  --epochs EPOCHS    upper epoch limit
-  --batch_size N     batch size
-  --bptt BPTT        sequence length
-  --dropout DROPOUT  dropout applied to layers (0 = no dropout)
-  --decay DECAY      learning rate decay per epoch
-  --tied             tie the word embedding and softmax weights
-  --seed SEED        random seed
-  --cuda             use CUDA
-  --log-interval N   report interval
-  --save SAVE        path to save the final model
+  -h, --help                       show this help message and exit
+  --data DATA                      location of the data corpus
+  --model MODEL                    type of recurrent net (RNN_TANH, RNN_RELU, LSTM, GRU)
+  --emsize EMSIZE                  size of word embeddings
+  --nhid NHID                      number of hidden units per layer
+  --nlayers NLAYERS                number of layers
+  --lr LR                          initial learning rate
+  --clip CLIP                      gradient clipping
+  --epochs EPOCHS                  upper epoch limit
+  --batch_size N                   batch size
+  --bptt BPTT                      sequence length
+  --dropout DROPOUT                dropout applied to layers (0 = no dropout)
+  --decay DECAY                    learning rate decay per epoch
+  --tied                           tie the word embedding and softmax weights
+  --seed SEED                      random seed
+  --cuda                           use CUDA
+  --log-interval N                 report interval
+  --save SAVE                      path to save the final model
+  --transformer_head N             the number of heads in the encoder/decoder of the transformer model
+  --transformer_encoder_layers N   the number of layers in the encoder of the transformer model
+  --transformer_decoder_layers N   the number of layers in the decoder of the transformer model
+  --transformer_d_ff N             the number of nodes on the hidden layer in feed forward nn
 ```
 
 With these arguments, a variety of models can be tested.
--- a/word_language_model/README.md
+++ b/word_language_model/README.md
@@ -4,8 +4,6 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) on a language modeli
 By default, the training script uses the Wikitext-2 dataset, provided.
 The trained model can then be used by the generate script to generate new text.
 
-The recent updates support a Transformer model, which is composed of encoder and decoder with multihead attention layers.
-
 ```bash 
 python main.py --cuda --epochs 6           # Train a LSTM on Wikitext-2 with CUDA
 python main.py --cuda --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA
--- a/word_language_model/README.md
+++ b/word_language_model/README.md
@@ -13,6 +13,8 @@ python main.py --cuda --epochs 6 --model Transformer --lr 5
                                            # Train a Transformer model on Wikitext-2 with CUDA
 python main.py --cuda --tied               # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs
 python generate.py                         # Generate samples from the trained LSTM model.
+python generate.py --cuda --model Transformer
+                                           # Generate samples from the trained Transformer model.
 ```
 
 The model uses the `nn.RNN` module (and its sister modules `nn.GRU` and `nn.LSTM`)
--- a/word_language_model/generate.py
+++ b/word_language_model/generate.py
@@ -57,14 +57,13 @@ if type(model).__name__ == "TransformerSeq2Seq":
     with open(args.outf, 'w') as outf:
         with torch.no_grad():  # no tracking history
             for i in range(args.words):
-                output = model(input, last_word, has_mask=False)
-                word_weights = output.squeeze().div(args.temperature).exp().cpu()
+                output = model(input, has_mask=False)
+                word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
                 word_idx = torch.multinomial(word_weights, 1)[0]
                 word = corpus.dictionary.idx2word[word_idx]
                 word_tensor = torch.ones(1).long().view(1,1).to(device)
                 word_tensor = word_tensor * word_idx
                 input = torch.cat([input, word_tensor], 0)
-                last_word = input[-1, :]
 
                 outf.write(word + ('\n' if i % 20 == 19 else ' '))
 
--- a/word_language_model/generate.py
+++ b/word_language_model/generate.py
@@ -15,53 +15,77 @@ parser = argparse.ArgumentParser(description='PyTorch Wikitext-2 Language Model'
 
 # Model parameters.
 parser.add_argument('--data', type=str, default='./data/wikitext-2',
-                    help='location of the data corpus')
+					help='location of the data corpus')
+parser.add_argument('--model', type=str, default='LSTM',
+					help='type of recurrent net (RNN_TANH, RNN_RELU, LSTM, GRU, Transformer)')
 parser.add_argument('--checkpoint', type=str, default='./model.pt',
-                    help='model checkpoint to use')
+					help='model checkpoint to use')
 parser.add_argument('--outf', type=str, default='generated.txt',
-                    help='output file for generated text')
+					help='output file for generated text')
 parser.add_argument('--words', type=int, default='1000',
-                    help='number of words to generate')
+					help='number of words to generate')
 parser.add_argument('--seed', type=int, default=1111,
-                    help='random seed')
+					help='random seed')
 parser.add_argument('--cuda', action='store_true',
-                    help='use CUDA')
+					help='use CUDA')
 parser.add_argument('--temperature', type=float, default=1.0,
-                    help='temperature - higher will increase diversity')
+					help='temperature - higher will increase diversity')
 parser.add_argument('--log-interval', type=int, default=100,
-                    help='reporting interval')
+					help='reporting interval')
 args = parser.parse_args()
 
 # Set the random seed manually for reproducibility.
 torch.manual_seed(args.seed)
 if torch.cuda.is_available():
-    if not args.cuda:
-        print("WARNING: You have a CUDA device, so you should probably run with --cuda")
+	if not args.cuda:
+		print("WARNING: You have a CUDA device, so you should probably run with --cuda")
 
 device = torch.device("cuda" if args.cuda else "cpu")
 
 if args.temperature < 1e-3:
-    parser.error("--temperature has to be greater or equal 1e-3")
+	parser.error("--temperature has to be greater or equal 1e-3")
 
 with open(args.checkpoint, 'rb') as f:
-    model = torch.load(f).to(device)
+	model = torch.load(f).to(device)
 model.eval()
 
 corpus = data.Corpus(args.data)
 ntokens = len(corpus.dictionary)
-hidden = model.init_hidden(1)
-input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
 
-with open(args.outf, 'w') as outf:
-    with torch.no_grad():  # no tracking history
-        for i in range(args.words):
-            output, hidden = model(input, hidden)
-            word_weights = output.squeeze().div(args.temperature).exp().cpu()
-            word_idx = torch.multinomial(word_weights, 1)[0]
-            input.fill_(word_idx)
-            word = corpus.dictionary.idx2word[word_idx]
+if args.model == "Transformer":
+	input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
 
-            outf.write(word + ('\n' if i % 20 == 19 else ' '))
+	with open(args.outf, 'w') as outf:
+		with torch.no_grad():  # no tracking history
+			for i in range(args.words):
+				seq_mask = model.generate_square_subsequent_mask(input.size(0)).to(device)
+				output = model(input, input, src_mask=seq_mask, tgt_mask=seq_mask)
+				word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
+				word_idx = torch.multinomial(word_weights, 1)[0]
+				word = corpus.dictionary.idx2word[word_idx]
+				word_tensor = torch.ones(1).long().view(1,1).to(device)
+				word_tensor = word_tensor * word_idx
+				input = torch.cat([input, word_tensor], 0)
 
-            if i % args.log_interval == 0:
-                print('| Generated {}/{} words'.format(i, args.words))
+				outf.write(word + ('\n' if i % 20 == 19 else ' '))
+
+				if i % args.log_interval == 0:
+					print('| Generated {}/{} words'.format(i, args.words))
+
+else:
+	hidden = model.init_hidden(1)
+	input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
+
+	with open(args.outf, 'w') as outf:
+		with torch.no_grad():  # no tracking history
+			for i in range(args.words):
+				output, hidden = model(input, hidden)
+				word_weights = output.squeeze().div(args.temperature).exp().cpu()
+				word_idx = torch.multinomial(word_weights, 1)[0]
+				input.fill_(word_idx)
+				word = corpus.dictionary.idx2word[word_idx]
+
+				outf.write(word + ('\n' if i % 20 == 19 else ' '))
+
+				if i % args.log_interval == 0:
+					print('| Generated {}/{} words'.format(i, args.words))
--- a/word_language_model/generate.py
+++ b/word_language_model/generate.py
@@ -51,12 +51,11 @@ corpus = data.Corpus(args.data)
 ntokens = len(corpus.dictionary)
 hidden = model.init_hidden(1)
 input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
-has_mask = False
 
 with open(args.outf, 'w') as outf:
     with torch.no_grad():  # no tracking history
         for i in range(args.words):
-            output, hidden = model(input, hidden, has_mask)
+            output, hidden = model(input, hidden, False)
             if model.model_type == 'Transformer':
                 word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
                 word_idx = torch.multinomial(word_weights, 1)[0]
--- a/word_language_model/generate.py
+++ b/word_language_model/generate.py
@@ -46,24 +46,28 @@ if args.temperature < 1e-3:
 with open(args.checkpoint, 'rb') as f:
     model = torch.load(f).to(device)
 model.eval()
+model.set_src_mask(None)
+model.set_tgt_mask(None)
+model.set_memory_mask(None)
 
 corpus = data.Corpus(args.data)
 ntokens = len(corpus.dictionary)
 
 if type(model).__name__ == "TransformerSeq2Seq":
     input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
+    last_word = input[-1, :]
 
     with open(args.outf, 'w') as outf:
         with torch.no_grad():  # no tracking history
             for i in range(args.words):
-                seq_mask = model.generate_square_subsequent_mask(input.size(0)).to(device)
-                output = model(input, input, src_mask=seq_mask, tgt_mask=seq_mask)
-                word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
+                output = model(input, last_word)
+                word_weights = output.squeeze().div(args.temperature).exp().cpu()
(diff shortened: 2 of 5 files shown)


zhangguanheng66:
A few minor updates. Not using superclass for the transformer model. Check transformer module exists before using. @cpuhrsch

word_language_model/README.md
```diff
@@ -21,24 +27,28 @@ The `main.py` script accepts the following arguments:
...
+  --batch_size N                   batch size
+  --bptt BPTT                      sequence length
+  --dropout DROPOUT                dropout applied to layers (0 = no dropout)
+  --decay DECAY                    learning rate decay per epoch
+  --tied                           tie the word embedding and softmax weights
+  --seed SEED                      random seed
+  --cuda                           use CUDA
+  --log-interval N                 report interval
+  --save SAVE                      path to save the final model
+  --transformer_head N             the number of heads in the encoder/decoder of the transformer model
```
cpuhrsch: Can some of this be meaningfully folded into the other arguments? Or put differently, can we recycle some of the existing arguments for the Transformer model? For example nlayers. Do we need separate flags for the encoder and decoder?

word_language_model/generate.py
```diff
@@ -49,19 +49,41 @@
...
-                print('| Generated {}/{} words'.format(i, args.words))
+
+if type(model).__name__ == "TransformerSeq2Seq":
+    input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
+
+    with open(args.outf, 'w') as outf:
+        with torch.no_grad():  # no tracking history
+            for i in range(args.words):
+                seq_mask = model.generate_square_subsequent_mask(input.size(0)).to(device)
+                output = model(input, input, src_mask=seq_mask, tgt_mask=seq_mask)
```
cpuhrsch: Since `seq_mask` is generated only to be fed into the model, maybe we can do this within the model forward?

word_language_model/generate.py
```diff
@@ -49,19 +49,41 @@
...
+                word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
```
cpuhrsch: Why do we need to index into the last element of output and not do that same with the LSTM?

word_language_model/main.py
```diff
@@ -53,6 +64,12 @@
     if not args.cuda:
         print("WARNING: You have a CUDA device, so you should probably run with --cuda")
 
+if args.model == 'Transformer':
+    try:
+        from torch.nn import Transformer
+    except:
+        raise ImportError('Transformer module exists in PyTorch 1.1 and above.')
```
cpuhrsch: I think it might be better to say "Transformer module does not exist in PyTorch 1.1 or lower".

word_language_model/main.py
```diff
@@ -92,7 +109,17 @@ def batchify(data, bsz):
 ###############################################################################
 
 ntokens = len(corpus.dictionary)
-model = model.RNNModel(args.model, ntokens, args.emsize, args.nhid, args.nlayers, args.dropout, args.tied).to(device)
+if args.model == 'Transformer':
```
cpuhrsch: Can we do this within Model? That is, you extend RNNModel to contain both and pass a flag to decide which one to use. This is in line with RNN settings GRU, LSTM, etc.

word_language_model/main.py
```diff
@@ -92,7 +109,17 @@ def batchify(data, bsz):
...
+	model = model.TransformerSeq2Seq(ntokens, ntokens, nhead=args.transformer_head,
```
cpuhrsch: RNNModel isn't using keyword arguments to pass ntokens, dropout etc. - We could be consistent so it's easier to read.

word_language_model/main.py
```diff
@@ -130,14 +157,24 @@ def evaluate(data_source):
...
+    else:
+        hidden = model.init_hidden(eval_batch_size)
     with torch.no_grad():
         for i in range(0, data_source.size(0) - 1, args.bptt):
             data, targets = get_batch(data_source, i)
-            output, hidden = model(data, hidden)
+
+            if args.model == 'Transformer':
+                if args.bptt > len(data_source) - 1 - i:
+                    tgt_mask = model.transformer.generate_square_subsequent_mask(len(data_source) - 1 - i).to(device)
```
cpuhrsch: Is there any way to do this implicitly within model and cache it via say `self.tgt_mask`?

word_language_model/main.py
```diff
@@ -147,14 +184,25 @@ def train():
...
         data, targets = get_batch(train_data, i)
         # Starting each batch, we detach the hidden state from how it was previously produced.
         # If we didn't, the model would try backpropagating all the way to start of the dataset.
-        hidden = repackage_hidden(hidden)
-        model.zero_grad()
-        output, hidden = model(data, hidden)
+
+        if args.model == 'Transformer':
+            if args.bptt > len(train_data) - 1 - i:
+                tgt_mask = model.transformer.generate_square_subsequent_mask(len(train_data) - 1 - i).to(device)
```
cpuhrsch: See above.

word_language_model/main.py
```diff
@@ -130,14 +157,24 @@ def evaluate(data_source):
     model.eval()
     total_loss = 0.
     ntokens = len(corpus.dictionary)
-    hidden = model.init_hidden(eval_batch_size)
+    if args.model == 'Transformer':
+        tgt_mask = model.transformer.generate_square_subsequent_mask(args.bptt).to(device)
```
cpuhrsch: Using member variables (`model.transformer`) directly is, in my opinion, not a good idea, because you might want to change them. You could either create a setter, or another method that relays it or hide this altogether as suggested in other comments.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+        num_encoder_layers: the number of sub-encoder-layers in the encoder (default=6).
+        num_decoder_layers: the number of sub-decoder-layers in the decoder (default=6).
+        dim_feedforward: the dimension of the feedforward network model (default=2048).
+        dropout: the dropout value (default=0.1).
+    Examples::
+        >>> transformer_model = nn.Transformer(src_vocab, tgt_vocab)
+        >>> transformer_model = nn.Transformer(src_vocab, tgt_vocab, nhead=16, num_encoder_layers=12)
+    """
+
+    def __init__(self, src_vocab, tgt_vocab, d_model=512, nhead=8, num_encoder_layers=6,
```
cpuhrsch: It could be cleaner to merge this into RNNModel, rename that to SeqModel and then have a flag that you can use to branch into each, as long as the forward methods can accept the same inputs and give the same outputs.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+        self._reset_parameters()
+
+    def _reset_parameters(self):
+        r"""Initiate parameters in the model."""
+
+        for p in self.parameters():
+            if p.dim() > 1:
+                nn.init.xavier_uniform_(p)
+
+    def forward(self, src, tgt, src_mask=None, tgt_mask=None, memory_mask=None):
```
cpuhrsch: Can these masks be cached via member variables?

word_language_model/main.py
zhangguanheng66: I was thinking about merging TransformerSeq2Seq and RNNModel, but realized they are very different in the architecture.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+        output = self.transformer.decoder(tgt, memory, tgt_mask, memory_mask)
+
+        if self.generator:
+            output = self.generator(output)
+
+        return output
+
+
+# Temporarily leave PositionalEncoding module here. Will be moved somewhere else.
+class PositionalEncoding(nn.Module):
```
cpuhrsch: I'd move this further up just because it's callsite is above it.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+        >>> pos_encoder = PositionalEncoding(d_model)
+    """
+
+    def __init__(self, d_model, dropout=0.1, max_len=5000):
+        super(PositionalEncoding, self).__init__()
+        self.dropout = nn.Dropout(p=dropout)
+
+        pe = torch.zeros(max_len, d_model)
+        position = torch.arange(0, max_len, dtype=torch.float).unsqueeze(1)
+        div_term = torch.exp(torch.arange(0, d_model, 2).float() * (-math.log(10000.0) / d_model))
```
cpuhrsch: What is the purpose of `-math.log(10000.0)`?

word_language_model/main.py
```diff
@@ -92,7 +109,17 @@ def batchify(data, bsz):
...
 ntokens = len(corpus.dictionary)
-model = model.RNNModel(args.model, ntokens, args.emsize, args.nhid, args.nlayers, args.dropout, args.tied).to(device)
+if args.model == 'Transformer':
+	model = model.TransformerSeq2Seq(ntokens, ntokens, nhead=args.transformer_head, 
+                                     d_model=args.emsize, dropout=args.dropout, 
+                                     num_encoder_layers=args.transformer_encoder_layers, 
+                                     num_decoder_layers=args.transformer_decoder_layers, 
+                                     dim_feedforward=args.transformer_dim_feedforward)
+	### Change the generator in transformer to nn.Linear(d_model, vocab)
+	model.generator = nn.Linear(args.emsize, ntokens)
```
cpuhrsch: Move this into the initialization. Replacing member variables after instantiation can cause pain further down the road, because you're not in control of your state.

word_language_model/README.md
```diff
@@ -4,11 +4,17 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) on a language modeli
...
-python main.py --cuda --epochs 6        # Train a LSTM on Wikitext-2 with CUDA
-python main.py --cuda --epochs 6 --tied # Train a tied LSTM on Wikitext-2 with CUDA
-python main.py --cuda --tied            # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs
-python generate.py                      # Generate samples from the trained LSTM model.
+The recent updates support a Transformer model, which is composed of encoder and decoder with multihead attention layers.
+
+```bash 
+python main.py --cuda --epochs 6           # Train a LSTM on Wikitext-2 with CUDA
+python main.py --cuda --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA
+python main.py --cuda --epochs 6 --model Transformer --lr 5
```
cpuhrsch: It could be good to add some expected numbers for the training results.

word_language_model/README.md
```diff
@@ -4,11 +4,17 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) on a language modeli
 By default, the training script uses the Wikitext-2 dataset, provided.
 The trained model can then be used by the generate script to generate new text.
 
-```bash
-python main.py --cuda --epochs 6        # Train a LSTM on Wikitext-2 with CUDA
-python main.py --cuda --epochs 6 --tied # Train a tied LSTM on Wikitext-2 with CUDA
-python main.py --cuda --tied            # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs
-python generate.py                      # Generate samples from the trained LSTM model.
+The recent updates support a Transformer model, which is composed of encoder and decoder with multihead attention layers.
```
cpuhrsch: I wouldn't mention that there was an update, just that the Transformer model exists, since at some point the update won't be recent anymore and then this needs to be updated.

word_language_model/main.py
cpuhrsch: Let's talk about this in person.

word_language_model/main.py
cpuhrsch: Move this into model.py

word_language_model/main.py
cpuhrsch: Should this be len(data) instead of len(data_source)?

word_language_model/main.py
```diff
@@ -130,14 +157,24 @@ def evaluate(data_source):
...
+        tgt_mask = model.transformer.generate_square_subsequent_mask(args.bptt).to(device)
+    else:
+        hidden = model.init_hidden(eval_batch_size)
     with torch.no_grad():
         for i in range(0, data_source.size(0) - 1, args.bptt):
             data, targets = get_batch(data_source, i)
-            output, hidden = model(data, hidden)
+
+            if args.model == 'Transformer':
+                if args.bptt > len(data_source) - 1 - i:
```
cpuhrsch: See if you can remove this, because it's throwing data away.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+                                          num_encoder_layers=num_encoder_layers,
+                                          num_decoder_layers=num_decoder_layers,
+                                          dim_feedforward=dim_feedforward, dropout=dropout)
+        self.src_embed = nn.Embedding(src_vocab, d_model)
+        self.pos_encoder = PositionalEncoding(d_model, dropout)
+        self.tgt_embed = nn.Embedding(tgt_vocab, d_model)
+        self.pos_decoder = PositionalEncoding(d_model, dropout)
+
+        self.generator = Generator(d_model, tgt_vocab)
+        self.src_vocab = src_vocab
```
cpuhrsch: We don't need to add state that we won't access, even if it's more complete.

word_language_model/main.py
cpuhrsch: You initialize model.generator with an instance of the Generator class, but then overwrite right after here. Also, generator is always set in this case and not a flag, so there's no need to branch on whether it exists or not within the forward method.

word_language_model/model.py
```diff
@@ -55,3 +57,165 @@ def init_hidden(self, bsz):
...
+        self.transformer = nn.Transformer(d_model=d_model, nhead=nhead,
+                                          num_encoder_layers=num_encoder_layers,
+                                          num_decoder_layers=num_decoder_layers,
+                                          dim_feedforward=dim_feedforward, dropout=dropout)
+        self.src_embed = nn.Embedding(src_vocab, d_model)
+        self.pos_encoder = PositionalEncoding(d_model, dropout)
+        self.tgt_embed = nn.Embedding(tgt_vocab, d_model)
+        self.pos_decoder = PositionalEncoding(d_model, dropout)
+
+        self.generator = Generator(d_model, tgt_vocab)
```
cpuhrsch: If this can be overwritten with nn.Linear you can get rid of the Generator class.

word_language_model/generate.py
zhangguanheng66: Generate seq_mask and tgt_mask is doable inside the forward function. Currently those are generated at the beginning of each epoch so they are re-used through the time steps. I thought this saves efforts.

word_language_model/generate.py
zhangguanheng66: AFAIK, in LSTM, the information from previous words are saving in the hidden and input is always a single word (here in this example). So the output from LSTM is also a single word. Differently in the transformer model, the information from previous words are carried by input. After each step, I attach the word_tensor to input by torch.cat. Therefore, the output is essentially a sentence. I dump the last word in the output to args.outf file.

word_language_model/main.py
zhangguanheng66: When we initiate the transformer model, we have no idea the length of tgt and src, unless we pass it through as an argument. The masks will change depending on the incoming tgt and src sequence. So it's better to generate them at the runtime.

word_language_model/model.py
zhangguanheng66: If those attributes are not used inside the transformer model, I will delete them to release the memory.

word_language_model/main.py
zhangguanheng66: It will not throwing data away. If the data left has a length short then args.bptt, it has to generate a new tgt_mask.

word_language_model/main.py
zhangguanheng66: No. This is for the case when the data left don't have a full length of args.bptt.

word_language_model/model.py
zhangguanheng66: The masks depends on the incoming src/tgt sequences.

word_language_model/model.py
zhangguanheng66: It's chosen by the paper "Attention is all you need." It forms a longer enough sin/cos wave length.

zhangguanheng66: 10000.0 is a random number. It is supposed to be large enough to cover the length of the incoming tgt/src sequences.

word_language_model/generate.py
cpuhrsch: Is it ok to return the last word as the output or will it interfere with other use-sites?

cpuhrsch: As in, the last word by default instead of the entire sentence.

word_language_model/model.py
cpuhrsch: But it looks like they're regenerated every time if called outside of this as well. So then let's at least do that in here to hide it I'd say.

word_language_model/generate.py
zhangguanheng66: Let me run a test: feed last word to decoder and a sentence to encoder. If no difference, I will switch to the same setup as LSTM.

word_language_model/generate.py
```diff
@@ -46,22 +46,48 @@
...
+    last_word = input[-1, :]
+
+    with open(args.outf, 'w') as outf:
+        with torch.no_grad():  # no tracking history
+            for i in range(args.words):
+                output = model(input, last_word)
+                word_weights = output.squeeze().div(args.temperature).exp().cpu()
+                word_idx = torch.multinomial(word_weights, 1)[0]
+                word = corpus.dictionary.idx2word[word_idx]
+                word_tensor = torch.ones(1).long().view(1,1).to(device)
```
cpuhrsch: It seems like this is very similar to filling input with wordidx. Why is it necessary to do this multiplication etc. instead of filling input?

word_language_model/generate.py
```diff
@@ -46,22 +46,48 @@
...
+                word_tensor = word_tensor * word_idx
+                input = torch.cat([input, word_tensor], 0)
+                last_word = input[-1, :]
```
cpuhrsch: Keeping this last word around is akin to "hidden" for the RNN.

word_language_model/main.py
```diff
@@ -53,6 +58,12 @@
     if not args.cuda:
         print("WARNING: You have a CUDA device, so you should probably run with --cuda")
 
+if args.model == 'Transformer':
```
cpuhrsch: Let's move this over into model.py since it's not related to training.

word_language_model/main.py
```diff
@@ -92,7 +103,12 @@ def batchify(data, bsz):
 ###############################################################################
 
 ntokens = len(corpus.dictionary)
-model = model.RNNModel(args.model, ntokens, args.emsize, args.nhid, args.nlayers, args.dropout, args.tied).to(device)
+if args.model == 'Transformer':
+	model = model.TransformerSeq2Seq(ntokens, ntokens, args.emsize, args.transformer_head,
+                                     args.nlayers, args.nlayers, args.nhid, 
+                                     args.dropout).to(device) 
+else:
+	model = model.RNNModel(args.model, ntokens, args.emsize, args.nhid, args.nlayers, args.dropout, args.tied).to(device)
```
cpuhrsch: Now that the arguments largely match up, can we have a "SeqModel" (or such) class that has a field to which it assigns either the RNN or Transformer based on the value of "args.model"?

word_language_model/main.py
```diff
@@ -147,14 +179,31 @@ def train():
...
-        hidden = repackage_hidden(hidden)
-        model.zero_grad()
-        output, hidden = model(data, hidden)
+
+        if args.model == 'Transformer':
+            if args.bptt > len(train_data) - 1 - i:
+                mask = model.generate_square_subsequent_mask(len(train_data) - 1 - i).to(device)
+                model.set_src_mask(mask)
+                model.set_tgt_mask(mask)
+                model.set_memory_mask(mask)
```
cpuhrsch: Why not do all of this within the forward pass?

word_language_model/model.py
cpuhrsch: Can't this be factored out to live outside of torch.exp? First multiplying by a very large negative number and then applying exp probably yields numerical instability.

word_language_model/model.py
```diff
@@ -55,3 +58,173 @@ def init_hidden(self, bsz):
...
+        Examples:
+            >>> output = pos_encoder(x)
+        """
+
+        x = x + self.pe[:x.size(0), :]
+        return self.dropout(x)
+
+
+# Temporarily leave Generator module here. Will be moved somewhere else.
+class Generator(nn.Module):
```
cpuhrsch: Do we still want to remove this?

zhangguanheng66: It's still used in the model.

word_language_model/main.py
zhangguanheng66: let's discuss this and make a decision.

word_language_model/model.py
cpuhrsch: But this is quite minimal. I think it's ok to merge it into TransformerSeq2Seq.

word_language_model/main.py
```diff
@@ -9,12 +9,13 @@
 
 import data
 import model
+import numpy as np
```
cpuhrsch: Don't add numpy as a dependency

word_language_model/model.py
zhangguanheng66: move to forward function

word_language_model/main.py
```diff
@@ -217,7 +227,9 @@ def export_onnx(path, batch_size, seq_len):
     model = torch.load(f)
     # after load the rnn params are not a continuous chunk of memory
     # this makes them a continuous chunk, and will speed up forward pass
-    model.rnn.flatten_parameters()
+    # Currently, only rnn model supports flatten_parameters function.
+    if args.model != 'Transformer':
```
cpuhrsch: I'd check explicitly for the models you want to allow this for. If someone ads a model this will cause a bug. It's a minor thing, but I think good practice.

word_language_model/model.py
```diff
@@ -1,57 +1,165 @@
...
-        if rnn_type in ['LSTM', 'GRU']:
-            self.rnn = getattr(nn, rnn_type)(ninp, nhid, nlayers, dropout=dropout)
+        self.model_type = model_type
+        self.nhead = nhead
+        self.ninp = ninp
+
+        if self.model_type in ['LSTM', 'GRU']:
+            self.model = getattr(nn, self.model_type)(ninp, nhid, nlayers, dropout=dropout)
+        elif self.model_type == 'Transformer':
+            try:
```
cpuhrsch: Move this to the top, since it's at a module level.

word_language_model/model.py
```diff
@@ -1,57 +1,165 @@
...
-        # "Using the Output Embedding to Improve Language Models" (Press & Wolf 2016)
-        # https://arxiv.org/abs/1608.05859
-        # and
-        # "Tying Word Vectors and Word Classifiers: A Loss Framework for Language Modeling" (Inan et al. 2016)
-        # https://arxiv.org/abs/1611.01462
-        if tie_weights:
-            if nhid != ninp:
-                raise ValueError('When using the tied flag, nhid must be equal to emsize')
-            self.decoder.weight = self.encoder.weight
+        if self.model_type == 'Transformer':
```
cpuhrsch: It should be neater to branch within "init_weights" instead of having a separate function. You can assert that tie_weights isn't used together with transformer.

word_language_model/model.py
```diff
@@ -1,57 +1,165 @@
...
-        self.nlayers = nlayers
+    def _reset_parameters(self):
+        r"""Initiate parameters in the model."""
+        assert self.model_type == 'Transformer'
+
+        for p in self.parameters():
+            if p.dim() > 1:
+                nn.init.xavier_uniform_(p)
+
+    def _set_src_mask(self, mask):
```
cpuhrsch: This only has two call sites right next to each other. I think you can inline it and get rid of it.

word_language_model/model.py
```diff
@@ -1,57 +1,165 @@
...
-        if self.rnn_type == 'LSTM':
+        if self.model_type == 'LSTM':
             return (weight.new_zeros(self.nlayers, bsz, self.nhid),
                     weight.new_zeros(self.nlayers, bsz, self.nhid))
         else:
             return weight.new_zeros(self.nlayers, bsz, self.nhid)
+
+
+# Temporarily leave PositionalEncoding module here. Will be moved somewhere else.
+class PositionalEncoding(nn.Module):
```
cpuhrsch: Now this needs to be moved above the Seq2Seq model haha

word_language_model/generate.py
```diff
@@ -51,14 +51,22 @@
 ntokens = len(corpus.dictionary)
 hidden = model.init_hidden(1)
 input = torch.randint(ntokens, (1, 1), dtype=torch.long).to(device)
+has_mask = False
```
cpuhrsch: You can just pass this in directly, since it's not an argument. If you want to document it more, you can pass it as a keyword argument.

word_language_model/generate.py
```diff
@@ -51,14 +51,22 @@
...
-            output, hidden = model(input, hidden)
-            word_weights = output.squeeze().div(args.temperature).exp().cpu()
-            word_idx = torch.multinomial(word_weights, 1)[0]
-            input.fill_(word_idx)
+            output, hidden = model(input, hidden, has_mask)
+            if model.model_type == 'Transformer':
+                word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
+                word_idx = torch.multinomial(word_weights, 1)[0]
+                word_tensor = torch.Tensor([[word_idx]]).long().to(device)
+                input = torch.cat([input, word_tensor], 0)
```
cpuhrsch: Doesn't this mean that input keeps on growing larger and larger? Is that actually necessary?

word_language_model/generate.py
```diff
@@ -51,14 +51,22 @@
...
 with open(args.outf, 'w') as outf:
     with torch.no_grad():  # no tracking history
         for i in range(args.words):
-            output, hidden = model(input, hidden)
-            word_weights = output.squeeze().div(args.temperature).exp().cpu()
-            word_idx = torch.multinomial(word_weights, 1)[0]
-            input.fill_(word_idx)
+            output, hidden = model(input, hidden, has_mask)
+            if model.model_type == 'Transformer':
+                word_weights = output[-1].squeeze().div(args.temperature).exp().cpu()
```
cpuhrsch: Now that we use an nn.Linear decoder I'm surprised you still need to retrieve the last entry.

word_language_model/model.py
```diff
@@ -1,57 +1,165 @@
...
+                if self.src_mask is None or self.src_mask.size(0) != len(input):
+                    mask = self._generate_square_subsequent_mask(len(input)).to(device)
+                    self._set_src_mask(mask)
+            else:
+                self._set_src_mask(None)
+
+            emb = self.drop(self.encoder(input))
+            output = emb * math.sqrt(self.ninp)
+            output = self.pos_encoder(output)
+            output = self.model(output, self.src_mask)
```
cpuhrsch: I'm confused about why we still need this mask, if the decoder is a Linear layer?

word_language_model/generate.py
zhangguanheng66: Not necessary. We can carry a constant length for the input sequence.

PR #556, opened 2019-05-08
Consistent learning rate for ResNet in readme

BradMcDanel:
The first example shows ResNet using the default learning rate (lr=0.1). However, the multiple GPU and multi-node examples show a lr=0.01 (which is the suggested learning rate for VGG and AlexNet). I assume this is a typo.


soumith:
cc: @ailzhang

Issue #559, opened 2019-05-13
GoogleNet does not work on imagenet


sakaia:
GoogLeNet does not run on imagenet training example.
Since googlenet model have multiple (logits etc) outputs.

To reproduce, do following instruction on Google Colab
'''
!wget https://s3.amazonaws.com/fast-ai-imageclas/imagenette-320.tgz
!tar xzf imagenette-320.tgz`
!git clone http://github.com/pytorch/examples
!pip uninstall -y torchvision
!git clone http://github.com/pytorch/vision
!cd vision;python setup.py install
!python examples/imagenet/main.py -a googlenet -j 2 -b 64 --epochs 10 imagenette-320/
'''

See following discussion
https://discuss.pytorch.org/t/question-about-googlenet/44896

msaroufim:
We stopped supporting googlenet since it's behavior is funky

emerth:
Hi msaroufim - sorry to revive an ancient thread. Is googlenet not supported only on ROCm, or not supported on CUDA as well?

PR #560, opened 2019-05-13
Fix googlenet

sakaia:
GoogLeNet model outputs 3 arguments.
This fix takes first tensor (logits).

Pull Request resuolved: #559


soumith:
googlenet is weird enough, i think we shouldn't support it in examples/imagenet/main.py 
the network doesn't converge when trained from scratch either

PR #565, opened 2019-05-27
Consider PyTorch 1.1 for _download_url_to_file module path

sakaia:
PyTorch #18758 moves _download_utl_to_file module path.
This patch considers it


soumith:
thank you @sakaia !

PR #569, opened 2019-06-01
incorrect data parallelism

andrewnaguib:
removed `.cuda()` from `DataParallel` init since `.cuda()` moves it to a single GPU device


andrewnaguib:
I will close this PR until I assure its correctness.

PR #583, opened 2019-07-03
C++ custom dataset

ShahriarSS:
Here is an example of loading custom dataset with OpenCV. The dataset used here is Caltec101. 
@yf225 @gchanan


cpp/custom-dataset/README.md
```diff
@@ -0,0 +1,52 @@
+# Custom Dataset Example with the PyTorch C++ Frontend
+
+This folder contains an example of loading a custom image dataset with OpenCV and training a model to label images, using the PyTorch C++ frontend.
```
yf225: Is there a way to provide instructions for installing OpenCV correctly? There seems to be multiple ways to do it, and having one that's guaranteed to work with this example would be beneficial. (We can have a separate README to document this.)

ShahriarSS: I will write a separate README for it.

ShahriarSS: @yf225 I have always installed OpenCV from package manager on Arch and Fedora for normal usage. It is hard to write instructions that work every where. Even the official documentation of OpenCV focuses on building from source. I did what I could. Please see if this is enough.

cpp/custom-dataset/README.md
```diff
@@ -0,0 +1,54 @@
...
+```shell
+$ cd custom-dataset
+$ mkdir build
+$ cd build
+$ cmake -DCMAKE_PREFIX_PATH=/path/to/libtorch ..
+$ make
+```
+
+where /path/to/libtorch should be the path to the unzipped LibTorch distribution, which you can get from the [PyTorch homepage](https://pytorch.org/get-started/locally/).
+
```
yf225: The LibTorch nightly or major release distribution might not be using a compiler that's ABI-compatible with the compiler used for building OpenCV. I think we should add a note here saying that if the user sees an error like `undefined reference to cv::imread(std::string const&, int)` when running the `make` command, they should build LibTorch from source using the instructions at https://github.com/pytorch/pytorch#from-source, and then set `CMAKE_PREFIX_PATH` to that PyTorch source directory.

yf225: The LibTorch nightly or major release distribution might not be using a compiler that's ABI-compatible with the compiler used for building OpenCV. I think we should add a note here saying that if the user sees an error like `undefined reference to cv::imread(std::string const&, int)` when running the `make` command, they should build LibTorch from source using the instructions at https://github.com/pytorch/pytorch#from-source, and then set `CMAKE_PREFIX_PATH` to that PyTorch source directory.

ShahriarSS:
You're welcome. I'll add more.

yf225:
@soumith This looks great to me and is ready to merge.

zirid:
Thanks for sharing this example.

I tried mnist example to check the instalation ; everything works fine but when i run your example i get an error (exception ) in training part at :

`auto loss = torch::nll_loss(output, batch.target);`  

Here is the  Prediction & ground_truth  output -- >
output

PR #585, opened 2019-07-04
Divide Up args.workers by ngpus_per_node

wangxicoding:
DIVUP args.workers by ngpus_per_node to make sure final workers > 0


soumith:
thanks!

PR #591, opened 2019-07-12
add a script to run all pytorch examples

nairbv:


run_python_examples.sh
```diff
@@ -0,0 +1,184 @@
...
+  EXAMPLE=${FUNCNAME[1]}
+  cd $BASE_DIR/$EXAMPLE
+  echo "Running example: $EXAMPLE"
+}
+
+function dcgan() {
+  start
+  if [ ! -d "lsun" ]; then
+    echo "cloning repo to get lsun dataset"
+    git clone https://github.com/fyu/lsun || { error "couldn't clone lsun repo needed for dcgan";  return; }
```
soumith: stuff like LSUN and Imagenet are really large datasets.
I wonder if we can do an environment variable like `$DATA_ROOT` to go look for them over there first. This way we can cache / mount them in CI

PR #596, opened 2019-07-18
Update README.md for Arguments

pushkalkatara:


soumith:
thanks!

pushkalkatara:
Hi @soumith, do we require support for [celebA](http://mmlab.ie.cuhk.edu.hk/projects/CelebA.html) dataset in the examples?

PR #602, opened 2019-08-06
Add an option to use multithreading to do mnist hogwild training

bddppq:


msaroufim:
3 years and no CLA check so closing

PR #610, opened 2019-08-14
Clean readme

SethHWeidman:
Added links to each of the examples in the README, and added a couple bullets for examples that weren't mentioned. 

This is the first part of the `examples` clean-up that will be part of the `tutorials` re-write.


soumith:
thanks!

PR #614, opened 2019-08-19
Fix 'RNNModel' object has no attribute 'model_type'

joistick11:
Model may not have attribute 'model_type' (in case of RNNModel)


soumith:
thank you!

vinutah:
thanks you @soumith @joistick11

PR #616, opened 2019-08-21
Reflecting the change in VAE paper name

JosephKJ:
The authors had renamed the paper to 'Auto-Encoding Variational Bayes'. This is more familiar to the current research community. I have just renamed the README to its current title, which matches their ICLR 2013 version.


soumith:
thanks!

PR #626, opened 2019-08-30
Added Transfer Learning using Libtorch and OpenCV

krshrimali:
This PR adds an example of Transfer Learning using `Libtorch` and `OpenCV`.

### DIFF
--- a/cpp/transfer-learning/classify.cpp
+++ b/cpp/transfer-learning/classify.cpp
@@ -1,0 +1,82 @@
+//
+//  classify.cpp
+//  transfer-learning
+//
+//  Created by Kushashwa Ravi Shrimali on 15/08/19.
+//  Copyright © 2019 Kushashwa Ravi Shrimali. All rights reserved.
+//
+
+#include <iostream>
+#include <torch/torch.h>
+#include <opencv2/opencv.hpp>
+#include <torch/script.h>
+#include <dirent.h>
+
+// Utility function to load image from given folder
+// File type accepted: .jpg
+std::vector<std::string> load_images(std::string folder_name) {
+    std::vector<std::string> list_images;
+    std::string base_name = folder_name;
+    DIR* dir;
+    struct dirent *ent;
+    if((dir = opendir(base_name.c_str())) != NULL) {
+        while((ent = readdir(dir)) != NULL) {
+            std::string filename = ent->d_name;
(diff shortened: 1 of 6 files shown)


cpp/transfer-learning/classify.cpp
```diff
@@ -0,0 +1,82 @@
+//
+//  classify.cpp
+//  transfer-learning
+//
+//  Created by Kushashwa Ravi Shrimali on 15/08/19.
+//  Copyright © 2019 Kushashwa Ravi Shrimali. All rights reserved.
+//
```
yf225: Do you think it would be possible to remove the copyright notice here? I looked at other examples in this repo and they don't seem to have copyright notices. Also having the notice might discourage people from using / modifying the code.

cpp/transfer-learning/classify.cpp
```diff
@@ -0,0 +1,82 @@
...
+    std::string location = argv[1];
+    
+    // argv[2] contains pre-trained model without last layer
+    // argv[3] contains trained last FC layer
+    std::string model_path = argv[2];
+    std::string model_path_linear = argv[3];
+    
+    // Load the model.
+    torch::jit::script::Module model;
+    model = torch::jit::load(model_path);
```
yf225: nit: these two lines can be combined into one `auto model = torch::jit::load(model_path)` if you prefer

krshrimali: Done! Thanks, @yf225 - I had included the type for users to know the type of model loaded by jit module. Added the type as a comment and used `auto model` now. Thanks!

cpp/transfer-learning/classify.cpp
krshrimali: Hi, sorry I missed this. This was automatically put by the IDE I used and I forgot to remove it. Thanks! :)

cpp/transfer-learning/classify.cpp
yf225: hey @krshrimali, sorry it was my mistake - you are right that writing `torch::jit::script::Module model = torch::jit::load(model_path)` is better because the user would have a clearer idea of the type of the model. Please feel free to change it back or have them written out in one line as you see fit :D

krshrimali: :D Thanks. @yf225 - I've committed the required changes. Please let me know if there are any further changes required.

yf225:
@soumith This looks great and is ready to merge. Thanks!

PR #629, opened 2019-09-10
Fix load model when a single GPU is used

cgnorthcutt:
This change is necessary for --resume and --evaluate methods when --gpu is specified. For example, if the model was trained on GPU 3 (on a machine with four GPUs: 0, 1, 2, and 3) and then you try to evaluate it on a machine with only 3 GPUs, you would get this error: 

RuntimeError: Attempting to deserialize object on CUDA device 3 but torch.cuda.device_count() is 3. Please use torch.load with map_location to map your storages to an existing device.

This change allows --evaluate and --resume functionalities to work correctly when a single GPU is used.


soumith:
thanks!

PR #654, opened 2019-11-02
Fix formatting to comply with PEP8

fshabashev:
Just a small formatting fix for mnist/main.py to make the code compliant  with the PEP8 styleguide.


soumith:
thanks!

PR #662, opened 2019-11-13
Adding Distributed Model Parallel Training Example for RNN

mrshenli:
The RNNModel puts the embedding table and the decoder parameters on
a remote parameter server, and locally holds parameters for the LSTM
module. The RNNModel overrides the forward() function to pass input
and intermediate outputs across workers. And it also added a
parameter_rrefs to return a list of RRefs of the model parameters.


mrshenli:
This PR depends on https://github.com/pytorch/pytorch/pull/29634

mrshenli:
cc @soumith

PR #664, opened 2019-11-17
Introduce --no-cuda option into snli per other examples.

dmiyakawa:
`mnist` has the same option and others often have similar capability like `--cuda`.

### DIFF
--- a/snli/train.py
+++ b/snli/train.py
@@ -14,7 +14,7 @@ from util import get_args, makedirs
 
 
 args = get_args()
-if torch.cuda.is_available():
+if not args.no_cuda and torch.cuda.is_available():
     torch.cuda.set_device(args.gpu)
     device = torch.device('cuda:{}'.format(args.gpu))
 else:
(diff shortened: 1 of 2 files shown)


msaroufim:
Not a bad addition per se but it only shows up in two places

```
(base) ➜  examples git:(main) grep -r "args.no_cuda" .
./mnist/main.py:    use_cuda = not args.no_cuda and torch.cuda.is_available()
./vae/main.py:args.cuda = not args.no_cuda and torch.cuda.is_available()
```

I'm thinking the more general ask is an interface for appropriate CLI arguments that people can use as a starting point

msaroufim:
Not a bad addition per se but it only shows up in two places

```
(base) ➜  examples git:(main) grep -r "args.no_cuda" .
./mnist/main.py:    use_cuda = not args.no_cuda and torch.cuda.is_available()
./vae/main.py:args.cuda = not args.no_cuda and torch.cuda.is_available()
```

I'm thinking the more general ask is an interface for appropriate CLI arguments that people can use as a starting point

PR #667, opened 2019-11-21
MNIST Example: Correct typo in default value within help

JamesOwers:
Tiny change, simply fix typo in the help string for --epochs.


soumith:
thanks!

PR #677, opened 2019-12-08
Implementing learning rate decay (a missing optional argument)

RikoNyberg:
Imitated the lr_decay implementation from [torch.optim.Adagrad](https://pytorch.org/docs/stable/_modules/torch/optim/adagrad.html#Adagrad)


msaroufim:
Been 3 years since CLA check not signed, closing

PR #678, opened 2019-12-10
update CXX compiler from 11 to 14

lzxzy:
According to [Tutorial Error for "Using the PyTorch C++ Frontend" #31037](https://github.com/pytorch/pytorch/issues/31037)
The latest libtorch version has been using some cpp 14 features. If we use cpp 11 for compile, it may import some errors like this `error: #error You need C++14 to compile PyTorch`
So I'm trying to update these demos.

### DIFF
--- a/cpp/dcgan/CMakeLists.txt
+++ b/cpp/dcgan/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 
 add_executable(dcgan dcgan.cpp)
 target_link_libraries(dcgan "${TORCH_LIBRARIES}")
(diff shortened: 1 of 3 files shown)


peterjc123:
cc @soumith

soumith:
cc: @yf225 @smessmer

I am not sure if I should merge this right now, or wait for the 14 transition to finish

smessmer:
The 14 transition is finished, we do require c++14 now. This can be merged.

soumith:
thanks a ton @lzxzy !

PR #683, opened 2019-12-23
Added continuous Bernoulli log normalizing constant to VAE

gabloa:
We recently had a paper at NeurIPS 2019 (https://arxiv.org/abs/1907.06845 and https://papers.nips.cc/paper/9484-the-continuous-bernoulli-fixing-a-pervasive-error-in-variational-autoencoders) where we show that when using a binary cross entropy reconstruction term in a VAE when the data is not binarized, a correction term (the log normalizing constant of a distribution which we call continuous Bernoulli) should be included. This pull request implements this correction on the VAE example. Resulting samples are significantly sharper and the resulting VAE has a proper probabilistic interpretation.

### DIFF
--- a/vae/main.py
+++ b/vae/main.py
@@ -7,7 +7,6 @@ from torch.nn import functional as F
 from torchvision import datasets, transforms
 from torchvision.utils import save_image
 
-
 parser = argparse.ArgumentParser(description='VAE MNIST Example')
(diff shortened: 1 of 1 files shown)


msaroufim:
It's an interesting contribution for sure but I'm think it's best to stick boring methods that work, I can't imagine this repo will be maintainable if we want to add all state of the art work in it

PR #684, opened 2019-12-24
change regression/poly_desc()

duguanyue:
In `regression/main.py`, the method `poly_desc()` creates a string description of a polynomial.
IMO, the polynomial coefficients should be from 1 up to 4, not 4 to 1, because `make_features()` method create a matrix with columns `[x, x^2, x^3, x^4]`.


soumith:
thanks!

PR #692, opened 2020-01-08
[WIP] Adding Distributed RL Model Example

mrshenli:
In this example, the RL model is distributed across one agent and
multiple observers. Each observer has a replicated submodel of
the RL model which all connect to the submodel on the agent.

During training, this example uses distributed autograd to set
gradients for all submodels. Then, it uses RPC calls to collect
gradients from all observers to the agent, sums those gradients,
applies the gradient to the local dummy model on agent, and then
broadcast the model parameters back to the observers to update
their models.

### DIFF
--- a/distributed/rpc/rl/distributed_model.py
+++ b/distributed/rpc/rl/distributed_model.py
@@ -1,0 +1,287 @@
+import argparse
+import gym
+import numpy as np
+import os
+from itertools import chain, count
+
+import torch
+import torch.distributed.autograd as dist_autograd
+import torch.distributed.rpc as rpc
+import torch.multiprocessing as mp
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from torch.distributed.rpc import RRef, rpc_sync, rpc_async, remote
+from torch.distributions import Categorical
+
+TOTAL_EPISODE_STEP = 5000
+AGENT_NAME = "agent"
+OBSERVER_NAME = "observer{}"
+
+parser = argparse.ArgumentParser(description='PyTorch RPC RL example')
+parser.add_argument('--world-size', type=int, default=2, metavar='W',
+                    help='world size for RPC, rank 0 is the agent, others are observers')
+parser.add_argument('--gamma', type=float, default=0.99, metavar='G',
+                    help='discount factor (default: 0.99)')
+parser.add_argument('--seed', type=int, default=543, metavar='N',
+                    help='random seed (default: 543)')
+parser.add_argument('--log-interval', type=int, default=10, metavar='N',
+                    help='interval between training status logs (default: 10)')
+args = parser.parse_args()
+
+torch.manual_seed(args.seed)
+
+
+def _call_method(method, rref, *args, **kwargs):
+    r"""
+    a helper function to call a method on the given RRef
+    """
+    return method(rref.local_value(), *args, **kwargs)
+
+
+def _remote_method(method, rref, *args, **kwargs):
+    r"""
+    a helper function to run method on the owner of rref and fetch back the
+    result using RPC
+    """
+    args = [method, rref] + list(args)
+    return rpc_sync(rref.owner(), _call_method, args=args, kwargs=kwargs)
+
+
+def _async_remote_method(method, rref, *args, **kwargs):
+    args = [method, rref] + list(args)
+    return rpc_async(rref.owner(), _call_method, args=args, kwargs=kwargs)
+
+class ObserverPolicy(nn.Module):
+    def __init__(self):
+        super(ObserverPolicy, self).__init__()
+        self.affine1 = nn.Linear(4, 128)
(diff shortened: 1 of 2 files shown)


distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+        super(ObserverPolicy, self).__init__()
+        self.affine1 = nn.Linear(4, 128)
+        self.dropout = nn.Dropout(p=0.6)
+        self.affine2 = nn.Linear(128, 2)
+
+    def forward(self, x):
+        x = self.affine1(x)
+        x = self.dropout(x)
+        x = F.relu(x)
+        return self.affine2(x)
```
mrshenli: each observer applies four layers in the forward pass

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+class AgentPolicy(nn.Module):
+    def __init__(self):
+        super(AgentPolicy, self).__init__()
+
+        self.saved_log_probs = []
+        self.rewards = []
+
+    def forward(self, action_scores):
+        return F.softmax(action_scores, dim=1)
```
mrshenli: each agent only applies a softmax.

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+            if done:
+                break
+
+class Agent:
+    def __init__(self, world_size):
+        self.ob_rrefs = []
+        self.agent_rref = RRef(self)
+        self.rewards = {}
+        self.saved_log_probs = {}
+        self.ob_policy = ObserverPolicy()
```
mrshenli: The agent also creates a dummy ObserverPolicy so that it can use the same optimizer to update all model parameters. Note that this ObserverPolicy never participates in the forward or backward pass. It is only used to apply the summed gradients.

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # collect gradients from observers
+        futs = []
+        # should be able to optimize this loop with a gather
+        for ob_rref in self.ob_rrefs:
+            futs.append(_async_remote_method(Observer.get_gradients, ob_rref, ctx_id))
+
+        grads = [fut.wait() for fut in futs]
+        grads = [*zip(*grads)]
+        grads = [sum(grad) for grad in grads]
```
mrshenli: the above few lines just sums the grads from all observers.

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+        # instead of putting it into a separate ctx
+        # set grads for observer model
+        ob_params = list(self.ob_policy.parameters())
+        for g, p in zip(grads, ob_params):
+            p.grad = g
+
+        # set grads for agent model
+        ctx_grads = dist_autograd.get_gradients(ctx_id)
+        for p in self.policy.parameters():
+            p.grad = ctx_grads[p]
```
mrshenli: Then, we set the grad field for all parameters. (note that the ob_policy is a dummy model, which is only useful for grad updates)

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.optimizer.step()
+
+        ob_params = [p.data.detach() for p in self.ob_policy.parameters()]
+        # should be able to optimize this loop with a scatter
+        futs = []
+        for ob_rref in self.ob_rrefs:
+            futs.append(_async_remote_method(Observer.update_model, ob_rref, ob_params))
+        for fut in futs:
+            fut.wait()
```
mrshenli: params on both the dummy model and the AgentPolicy are updated. Now broadcast the dummy model params to all observers to perform updates there.

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+
```
mrshenli: I believe the above few steps can be replaced by c10d gather/scatter, and wrapped into your own version of distributed optimizer. In that way, it will be more efficient and look better.

distributed/rpc/rl/distributed_model.py
```diff
@@ -0,0 +1,287 @@
...
+class Observer:
+    def __init__(self):
+        self.id = rpc.get_worker_info().id
+        self.env = gym.make('CartPole-v1')
+        self.env.seed(args.seed)
+        self.policy = ObserverPolicy()
+
+    def get_gradients(self, ctx_id):
+        all_grads = dist_autograd.get_gradients(ctx_id)
```
mrshenli: After the distributed backward pass, the grads for the `ObserverPolicy` model lives in `dist_autograd.get_gradients(ctx_id)` on the observer. Retrieve them and give them to the agent.

PR #699, opened 2020-01-20
dcgan: using `fake` dataset still requires dataroot parameter

### LINKED ISSUE #698: dcgan fails on "fake" dataset
I have been using the dcgan example as a stress test for a machine. To save time, I have made use of the `fake` dataset.

`python main.py --dataset 'fake'` fails because `dataroot` is a required parameter. However, the `fake` dataset, does not need such information.

epignatelli:
Fixes #698 

#### To reproduce:

```sh
cd dcgan
python main.py --dataset fake
```


soumith:
thanks!

PR #707, opened 2020-02-08
Add help message for snli example

limkeunhak:


soumith:
thanks, this is great!

Issue #706, opened 2020-02-08
Add new feature to show the result of prediction 


JieunKim4578:
I want to add new method to this mnist example. The motivation for this PR is to see the real result of prediction of this training model. Below is the simple blueprint of what i want to implement. Is it ok to submit these kind of method? 

```py

    def show_prediction(args, model, device, test_loader):
    model.eval()
    with torch.no_grad():
         for data, target in test_loader:
             data, target = data.to(device), target.to(device)
             plt.imshow(data[0].reshape(28, 28), cmap=cm.gray_r)
             plt.show()

             output = model(data)
             print("output[0]: ", output[0])

             pred = output.argmax(dim=1, keepdim=True)
             print("pred[0]: ", pred[0])
```

Example.
![image](https://user-images.githubusercontent.com/52270723/74081798-4916c400-4a96-11ea-9fa5-7045d2997709.png)
![image](https://user-images.githubusercontent.com/52270723/74081809-62b80b80-4a96-11ea-89d0-341b966abdbe.png)

msaroufim:
Since main PR was also closed

PR #708, opened 2020-02-08
Add --show-test option.

JieunKim4578:
Add 'show_test' method. 'show_test' will show the prediction result of mnist after training model.
It will display the image of 10 different handwriting number with the prediction number result.
To see the results, limit the test result only to first index.

Change default=False if you don't want to use this.

This PR is related to #706.


JieunKim4578:
This is the expected result. 
image

soumith:
i'm declining to merge this into the example, which needs to be kept concise as a first principle

PR #709, opened 2020-02-08
Update .gitignore

limkeunhak:
Add files that do not need tracking generated when running examples below.

- dcgan

- snli

- word_language_model

- fast_neural_style


soumith:
thanks!

PR #719, opened 2020-02-17
bug: remove log_softmax because criterion is CrossEntropyLoss

noowad93:


msaroufim:
Closing since no response from OP

PR #722, opened 2020-02-28
Reinforce Policy Update wrong

karanchahal:
The update step for REINFORCE appears to be wrong. Instead of multiplying log probs and rewards at each step, we should sum up all the rewards to go, and the log probs across the entire episode. After getting these sums, we should multiply these 2 values together. This was mentioned in Sergey Levine's lecture on Policy gradients where he derived the entire policy gradient algorithm.

Source: https://www.youtube.com/watch?v=Ds1trXd6pos&list=PLkFD6_40KJIwhWJpGazJ9VSj9CFMkb79A&index=5

### DIFF
--- a/reinforcement_learning/reinforce.py
+++ b/reinforcement_learning/reinforce.py
@@ -68,10 +68,9 @@ def finish_episode():
         returns.insert(0, R)
     returns = torch.tensor(returns)
     returns = (returns - returns.mean()) / (returns.std() + eps)
-    for log_prob, R in zip(policy.saved_log_probs, returns):
-        policy_loss.append(-log_prob * R)
-    optimizer.zero_grad()
-    policy_loss = torch.cat(policy_loss).sum()
+    l_probs = policy.saved_log_probs.sum()
+    rews = returns.sum()
+    policy_loss = rews + l_probs
     policy_loss.backward()
     optimizer.step()
     del policy.rewards[:]


sammiie:
To address the "Reinforce Policy Update wrong" issue #722, you will need to review and correct the code that is responsible for the policy update in your Reinforcement Learning implementation. Below, I'll provide a general code template that demonstrates how to perform a policy update using the REINFORCE algorithm. You should adapt this code to your specific RL environment and problem.

Import Necessary Libraries:
Import the required libraries, such as PyTorch or TensorFlow, if you're using a deep learning framework for your RL implementation.

Define Your Policy Network:
Define the neural network that represents your policy. This network will output action probabilities based on the current state.

import torch
import torch.nn as nn
import torch.optim as optim

# Define a simple policy network (example)
class PolicyNetwork(nn.Module):
    def __init__(self, input_size, output_size):
        super(PolicyNetwork, self).__init__()
        self.fc = nn.Linear(input_size, output_size)
        self.softmax = nn.Softmax(dim=-1)

    def forward(self, state):
        logits = self.fc(state)
        action_probs = self.softmax(logits)
        return action_probs


Initialize Policy and Optimizer:
Initialize your policy network and an optimizer (e.g., SGD or Adam).

# Initialize policy network and optimizer
policy_net = PolicyNetwork(input_size, output_size)
optimizer = optim.Adam(policy_net.parameters(), lr=learning_rate)


Policy Update (REINFORCE Algorithm):
Implement the policy update step using the REINFORCE algorithm. This typically involves the following steps:

Sample actions from the policy.
Compute the log probabilities of the sampled actions.
Compute the rewards.
Compute the loss using the log probabilities and rewards.
Perform gradient ascent using the loss to update the policy network.
Here's a simplified example:


def reinforce_update(policy_net, optimizer, rewards, log_probs):
    loss = -torch.mean(log_probs * rewards)

    optimizer.zero_grad()
    loss.backward()
    optimizer.step()



You will need to integrate this policy update code into your RL training loop, considering the specifics of your environment and problem.

Training Loop:
Implement the RL training loop where you interact with the environment, collect experiences, compute rewards, and update the policy based on the REINFORCE algorithm.



for episode in range(num_episodes):
    # Interaction with the environment and collecting experiences
    # Compute rewards for each time step

    # Compute log probabilities of actions taken

    # Perform REINFORCE policy update
    reinforce_update(policy_net, optimizer, rewards, log_probs)


Please adapt the code provided above to your specific RL environment and problem. Make sure to follow the best practices for reinforcement learning, including handling exploration, discount factors, and other relevant factors specific to your problem.

PR #731, opened 2020-03-11
loss detach

developer0hye:
loss detach


soumith:
thanks!

colesbury:
This doesn't do anything... the detach() is unnecessary

developer0hye:
@colesbury 
Yeah, I agree with you. I realized that after merge...

PR #732, opened 2020-03-12
fix loss calculation for RNN

taufique74:
After removal of the CrossEntropyLoss, loss calculation for RNN is broken.
Currently it's taking NLLLoss of direct RNN output (not the LogSoftmax), resulting in negative losses.

### DIFF
--- a/word_language_model/main.py
+++ b/word_language_model/main.py
@@ -145,10 +145,10 @@ def evaluate(data_source):
             data, targets = get_batch(data_source, i)
             if args.model == 'Transformer':
                 output = model(data)
+                output = output.view(-1, ntokens)
             else:
                 output, hidden = model(data, hidden)
                 hidden = repackage_hidden(hidden)
-            # output_flat = output.view(-1, ntokens)
             total_loss += len(data) * criterion(output, targets).item()
     return total_loss / (len(data_source) - 1)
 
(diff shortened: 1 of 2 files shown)


word_language_model/main.py
```diff
@@ -145,11 +145,11 @@ def evaluate(data_source):
             data, targets = get_batch(data_source, i)
             if args.model == 'Transformer':
                 output = model(data)
+                output = output.view(-1, ntokens)
             else:
                 output, hidden = model(data, hidden)
                 hidden = repackage_hidden(hidden)
-            output_flat = output.view(-1, ntokens)
-            total_loss += len(data) * criterion(output_flat, targets).item()
+            total_loss += len(data) * criterion(output, targets).item()
```
zhangguanheng66: Should it be `criterion(output_flat, targets)`?

taufique74: I modified the `RNNModel` so that it returns the `LogSoftMax` of flatten output. 
Since the transformer requires flattening, it was handled in the if condition. 

Although this can be done in other ways. It was a quick fix since it was producing negative losses for RNN models.
We can import `log_softmax` in `main.py` and add another if statement. If it's RNN, just adding `criterion(log_softmax(output_flat, dim=-1), targets)` would also do the job. 
`log_softmax` was already imported in `model.py`, so I didn't want to import in `main.py` too.

PR #733, opened 2020-03-12
[Dist Autograd - API Change] Updated dist_autograd and dist_optim to be functional

osalpekar:
`dist_autograd()` and `dist_optimizer.step` now explicitly take in the `context_id`. This PR reflects the API change in the RPC-based RNN training example.

These API changes will be a part of the PyTorch 1.5 Release.


mrshenli:
@nairbv Shall we land this PR for v1.5? Thanks!

PR #750, opened 2020-04-16
Add an exception for nan loss during training

developer0hye:
Add an exception for nan loss during training


msaroufim:
I'm not sure throwing an exception is the correct behavior since it's better to have the program running to know why nan happened - feel free to reopen if you disagree

PR #765, opened 2020-04-29
Update dcgan.cpp

kasimte:
Fix AdamOptions initialization error with latest libtorch.

Error without this change: 

```
dcgan.cpp:124:64: error: no member named 'beta1' in 'torch::optim::AdamOptions'
      generator->parameters(), torch::optim::AdamOptions(2e-4).beta1(0.5));
```

### DIFF
--- a/cpp/dcgan/dcgan.cpp
+++ b/cpp/dcgan/dcgan.cpp
@@ -121,9 +121,9 @@ int main(int argc, const char* argv[]) {
       torch::data::DataLoaderOptions().batch_size(kBatchSize).workers(2));
 
   torch::optim::Adam generator_optimizer(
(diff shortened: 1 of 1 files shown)


soumith:
thank you!

dayongt:
I use the cu10.1 debug libtorch on windows. Everything is ok when training on CPU. However, when training on GPU, each time the backward() method is called, it produces a warning:
`[W ..\..\aten\src\ATen\native\Resize.cpp:19] Warning: An output with one or more elements was resized since it had shape [64, 1, 1, 1], which does not match the required output shape [64, 1, 1, 64].This behavior is deprecated, and in a future PyTorch release outputs will not be resized unless they have zero elements. You can explicitly reuse an out tensor t by resizing it, inplace, to zero elements with t.resize_(0). (function resize_output)`

PR #767, opened 2020-05-03
Eliminate .data access for parameters as much as possible

vishwakftw:
Following https://github.com/pytorch/pytorch/pull/34211, this PR removes all direct accesses to the `.data` attribute.


soumith:
nice!

PR #774, opened 2020-05-21
Fix Parameter Server Example

osalpekar:
This fixes incorrect arguments passed to `model.forward` in the RPC Parameter Server training example.


jinalong:
@osalpekar Dear Osalpekar, I can corrently run parameter server on two nodes. However, when using three nodes, I get the following error:

jinalong:
@osalpekar Dear Osalpekar, I can run parameter server on two nodes without GPU. However, when running on three nodes, I got the following error: RuntimeError: Error on Node 0: one of the variables needed for gradient computation has been modified by an operation: [CPUFloatType [9216, 128]], which is output 0 of TBackward, is at version 5; expected version 4 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).

PR #775, opened 2020-05-21
[word_language_model] Fix initialization of weight of Decoder

manashmandal:
`nn.init.zeros_(self.decoder)` throws this error `AttributeError: module 'torch.nn.init' has no attribute 'zero_'` because the weights weren't passed.

### DIFF
--- a/word_language_model/model.py
+++ b/word_language_model/model.py
@@ -42,7 +42,7 @@ class RNNModel(nn.Module):
     def init_weights(self):
         initrange = 0.1
         nn.init.uniform_(self.encoder.weight, -initrange, initrange)
-        nn.init.zeros_(self.decoder)
+        nn.init.zeros_(self.decoder.weight)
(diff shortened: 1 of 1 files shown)


soumith:
thanks!

word_language_model/model.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, rnn_type, ntoken, ninp, nhid, nlayers, dropout=0.5, tie_weigh
     def init_weights(self):
         initrange = 0.1
         nn.init.uniform_(self.encoder.weight, -initrange, initrange)
-        nn.init.zeros_(self.decoder)
+        nn.init.zeros_(self.decoder.weight)
```
csukuangfj, line 45: Is there a point to invoke `nn.init.zeros_` here?
The next line re-initializes `self.decoder.weight`.

PR #779, opened 2020-05-25
[word_language_model] Fix grad in-place operation

yangchris11:
`p.add_(-lr, p.grad)` throws `RuntimeError: a leaf Variable that requires grad is being used in an in-place operation`, using `p.data.add_(-lr, p.grad)` fixes this issue.


soumith:
thanks!

YYStreet:
Please also fix this line https://github.com/pytorch/examples/blob/master/regression/main.py#L60.

Same issue

YYStreet:
> Please also fix this line https://github.com/pytorch/examples/blob/master/regression/main.py#L60.
> 
> Same issue

Create a PR https://github.com/pytorch/examples/pull/781

PR #781, opened 2020-05-28
fix the issue PR #779 left

YYStreet:
https://github.com/pytorch/examples/pull/779/files did not fix the issue present in regression test


soumith:
thank you.

PR #792, opened 2020-06-18
Adding Batch RPC example

mrshenli:
This PR adds two examples for `@rpc.functions.async_execution` decorator.
    
1. A synchronized batch-pdate Parameter Server example.
2. A batch-processing reinforcement learning example.

### DIFF
--- a/distributed/rpc/batch/parameter_server.py
+++ b/distributed/rpc/batch/parameter_server.py
@@ -1,0 +1,144 @@
+import os
+import threading
+from datetime import datetime
+
+import torch
+import torch.distributed.rpc as rpc
+import torch.multiprocessing as mp
+import torch.nn as nn
+from torch import optim
+
+import torchvision
+
+
+batch_size = 20
+image_w = 64
+image_h = 64
+num_classes = 30
+batch_update_size = 5
+num_batches = 6
+
+
+def timed_log(text):
+    print(f"{datetime.now().strftime('%H:%M:%S')} {text}")
+
+
+class BatchUpdateParameterServer(object):
+
+    def __init__(self, batch_update_size=batch_update_size):
+        self.model = torchvision.models.resnet50(num_classes=num_classes)
+        self.lock = threading.Lock()
+        self.future_model = torch.futures.Future()
+        self.batch_update_size = batch_update_size
+        self.curr_update_size = 0
+        self.optimizer = optim.SGD(self.model.parameters(), lr=0.001, momentum=0.9)
+        for p in self.model.parameters():
+            p.grad = torch.zeros_like(p)
+
+    def get_model(self):
+        return self.model
+
+    @staticmethod
+    @rpc.functions.async_execution
+    def update_and_fetch_model(ps_rref, grads):
+        self = ps_rref.local_value()
+        timed_log(f"PS got {self.curr_update_size}/{batch_update_size} updates")
+        for p, g in zip(self.model.parameters(), grads):
+            p.grad += g
+        with self.lock:
+            self.curr_update_size += 1
+            fut = self.future_model
+
+            if self.curr_update_size >= self.batch_update_size:
+                for p in self.model.parameters():
+                    p.grad /= self.batch_update_size
+                self.curr_update_size = 0
+                self.optimizer.step()
+                self.optimizer.zero_grad()
+                fut.set_result(self.model)
+                timed_log("PS updated model")
+                self.future_model = torch.futures.Future()
+
+        return fut
+
+
+class Trainer(object):
+
+    def __init__(self, ps_rref):
+        self.ps_rref = ps_rref
+        self.loss_fn = nn.MSELoss()
+        self.one_hot_indices = torch.LongTensor(batch_size) \
+                                    .random_(0, num_classes) \
+                                    .view(batch_size, 1)
+
+    def get_next_batch(self):
+        for _ in range(num_batches):
+            inputs = torch.randn(batch_size, 3, image_w, image_h)
+            labels = torch.zeros(batch_size, num_classes) \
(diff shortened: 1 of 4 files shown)


mrshenli:
The figure below shows the impact of `@rpc.functions.async_execution` for the reinforcement learning example, where batch processing significantly speeds up training.

batch

distributed/rpc/batch/requirements.txt
```diff
@@ -0,0 +1,3 @@
+torch
```
rohan-varma, line 1: Do we need to enforce at least v1.6.0 or should it automatically do that?

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __init__(self, batch_update_size=batch_update_size):
+        self.model = torchvision.models.resnet50(num_classes=num_classes)
+        self.lock = threading.Lock()
+        self.future_model = torch.futures.Future()
+        self.batch_update_size = batch_update_size
+        self.curr_update_size = 0
+        self.optimizer = optim.SGD(self.model.parameters(), lr=0.001, momentum=0.9)
+        for p in self.model.parameters():
+            p.grad = torch.zeros_like(p)
```
rohan-varma, line 40: Is `p.grad.zero_()` more canonical?

distributed/rpc/batch/requirements.txt
mrshenli: Good point, let me add a check to the code.

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,144 @@
...
+            rpc_backend_options=options
+        )
+        run_ps([f"trainer{r}" for r in range(1, world_size)])
+
+    # block until all rpcs finish
+    rpc.shutdown()
+
+
+if __name__=="__main__":
+    world_size = 6
```
rohan-varma, line 143: IIUC there needs to be at least `batch_update_size` trainers, otherwise, in the first iteration, the trainer will never get unblocked from the RPC that waits for 5 updates right? Should we code this value to `world_size = batch_update_size + 1` ?

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,144 @@
...
+            yield inputs.cuda(), labels.cuda()
+
+    def train(self):
+        name = rpc.get_worker_info().name
+        m = self.ps_rref.rpc_sync().get_model().cuda()
+        for inputs, labels in self.get_next_batch():
+            timed_log(f"{name} processing one batch")
+            self.loss_fn(m(inputs), labels).backward()
+            timed_log(f"{name} reporting grads")
+            m = rpc.rpc_sync(
```
rohan-varma, line 88: As a possible extension, is it possible for this trainer to continue training with local model until all the updates are done and then we update the model with the remote model asynchronously? Not sure what the implications of this is accuracy-wise, was just wondering if one method is preferred over the other.

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __init__(self, ps_rref):
+        self.ps_rref = ps_rref
+        self.loss_fn = nn.MSELoss()
+        self.one_hot_indices = torch.LongTensor(batch_size) \
+                                    .random_(0, num_classes) \
+                                    .view(batch_size, 1)
+
+    def get_next_batch(self):
+        for _ in range(6):
```
rohan-varma, line 75: I'm assuming this is unrelated to `batch_update_size` right? Can we have a constant for this such as `num_batches`?

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,144 @@
...
+    def get_model(self):
+        return self.model
+
+    @staticmethod
+    @rpc.functions.async_execution
+    def update_and_fetch_model(ps_rref, grads):
+        self = ps_rref.local_value()
+        with self.lock:
+            self.curr_update_size += 1
+            timed_log(f"PS got one update, {self.curr_update_size} in batch")
```
rohan-varma, line 46: Nit: Add `batch_update_size` for the logging information so users have an idea of how many more updates it's blocked on?

distributed/rpc/batch/parameter_server.py
```diff
@@ -0,0 +1,144 @@
...
+            for p, g in zip(self.model.parameters(), grads):
+                p.grad += g
```
rohan-varma, line 48: Would it be hogwild if we don't have the lock?

distributed/rpc/batch/parameter_server.py
mrshenli: `p.grad` is actually `None` initially. `self.optimizer.zero_grad()` also won't touch None grads.

distributed/rpc/batch/parameter_server.py
mrshenli: Good catch! Hard-coding these two args are bad. Let me update

distributed/rpc/batch/parameter_server.py
mrshenli: Yep, I believe that's possible and also would very likely to speed up training without too much accuracy penalty. But his example was intended to serve as a simple starter for batch-processing, so would like to keep both the concept and implementation straightforward. Let me update the README to explain this.

distributed/rpc/batch/parameter_server.py
mrshenli: Good point! I think this part can be moved out of the lock as the there will be only one thread running at a time due to GIL. Will update.

PR #749, opened 2020-04-12
Adding distributed pipeline parallelism example

mrshenli:
This example shows how to use RPC to implement pipeline parallelism. This can be viewed as a distributed version of [single machine multiple GPU pipeline parallelism](https://pytorch.org/tutorials/intermediate/model_parallel_tutorial.html#speed-up-by-pipelining-inputs).

The numbers below show how the total execution time decreases with the increase of `num_split`.

```
$ python main.py 
Processing batch 0
Processing batch 1
Processing batch 2
number of splits = 1, execution time = 16.45062756538391
Processing batch 0
Processing batch 1
Processing batch 2
number of splits = 2, execution time = 12.329529762268066
Processing batch 0
Processing batch 1
Processing batch 2
number of splits = 4, execution time = 10.164430618286133
Processing batch 0
Processing batch 1
Processing batch 2
number of splits = 8, execution time = 9.076049566268921
```

PR #796, opened 2020-07-01
Update pipeline parallel example to use RRef helpers

mrshenli:


mrshenli:
Addressed @osalpekar's comments in #749 as well. Sorry, I should have address those in the previous PR.

distributed/rpc/pipeline/main.py
```diff
@@ -248,6 +207,9 @@ def run_master(split_size):
         labels = torch.zeros(batch_size, num_classes) \
                       .scatter_(1, one_hot_indices, 1)
 
+        # The distributed autograd context is the deciated scope for the
```
osalpekar, line 210: nit: was this supposed to be "dedicated scope"?

mrshenli: good catch! Thanks!

PR #794, opened 2020-06-24
tweak running examples without cuda

mattip:
Tweak the `run_python_examples.sh` script to complete in ~5 minutes (on my machine) without CUDA. Once this is merged, we can consider adding #704 to actually run the examples as part of CI.

TBD: how do we get a pre-release version of torch/torchvision in order to test the latest HEAD? Perhaps it should be a pytorch/pytorch cron job that runs once a day or once a week?

ping @jlin27

PR #800, opened 2020-07-08
Example for combining DDP + RPC

pritamdamania87:
Summary: The example includes a simple model consisting of a sparse part
and a dense part. The sparse part is an nn.EmbeddingBag stored on a
parameter server and the dense part is an nn.Linear module residing on
the trainers. The dense part on the trainers are replicated via
DistributedDataParallel.

A master creates the nn.EmbeddingBag and drives the training loop on the
trainers. The training loop performs an embedding lookup via the
Distributed RPC Framework and then executes the local dense component.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


distributed/rpc/ddp_rpc/main.py
```diff
@@ -0,0 +1,183 @@
...
+                world_size=world_size,
+                rpc_backend_options=rpc_backend_options)
+        # parameter server do nothing
+        pass
+
+    # block until all rpcs finish
+    rpc.shutdown()
+
+
+if __name__=="__main__":
```
nairbv, line 180: can we add this to the run_python_examples.sh script?

distributed/rpc/ddp_rpc/README.md
```diff
@@ -0,0 +1,20 @@
+Distributed DataParallel + Distributed RPC Framework Example
+
+The example shows how to combine Distributed DataParallel with the Distributed 
+RPC Framework. There are two trainer nodes, 1 master node and 1 parameter in
```
mrshenli, line 4: parameter -> parameter server

distributed/rpc/ddp_rpc/main.py
```diff
@@ -0,0 +1,183 @@
...
+import torch
+import torch.distributed as dist
+import torch.distributed.autograd as dist_autograd
+import torch.distributed.rpc as rpc
+from torch.distributed.rpc import ProcessGroupRpcBackendOptions
+import torch.multiprocessing as mp
+import torch.optim as optim
+from torch.distributed.optim import DistributedOptimizer
+from torch.distributed.rpc import RRef
+from torch.nn.parallel import DistributedDataParallel as DDP
```
mrshenli, line 14: Let's sort the import block a bit.

distributed/rpc/ddp_rpc/main.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def forward(self, indices, offsets):
+        emb_lookup = self.emb_rref.rpc_sync().forward(indices, offsets)
+        return self.fc(emb_lookup.cuda(self.device))
+
+def _retrieve_embedding_parameters(emb_rref):
+    param_rrefs = []
+    for param in emb_rref.local_value().parameters():
+        param_rrefs.append(RRef(param))
+    return param_rrefs
```
mrshenli, line 44: can be abbreviated to the following?

```python
return [RRef(p) for p in emb_rref.local_value().parameters()]
```

distributed/rpc/ddp_rpc/main.py
```diff
@@ -0,0 +1,183 @@
...
+    r"""
+    A wrapper function that initializes RPC, calls the function, and shuts down
+    RPC.
+    """
+    os.environ['MASTER_ADDR'] = 'localhost'
+    os.environ['MASTER_PORT'] = '29500'
+
+
+    rpc_backend_options = ProcessGroupRpcBackendOptions()
+    rpc_backend_options.init_method='tcp://localhost:29501'
```
mrshenli, line 125: Will it be better to use init_method for both RPC and DDP to stay consistent?

mrshenli: Shall we add a comment here to mention that it needs to use a different port number as RPC will create a different process group?

distributed/rpc/ddp_rpc/main.py
pritamdamania87: It looks like `distributed` is commented out in that script: https://github.com/pytorch/examples/blob/master/run_python_examples.sh#L178?

nairbv: hmm... was just commented out a week ago, could have been by mistake:

https://github.com/pytorch/examples/pull/794

nairbv: oh, actually they were added as commented out. I think we can just add a distributed function and uncomment

mrshenli:
cc @jlin27

distributed/rpc/ddp_rpc/main.py
pritamdamania87: Looking at https://github.com/pytorch/examples/pull/794 it seems like the goal was to have the entire script run within 5 minutes. If we add distributed to it, I don't think we can satisfy the goal without updating other examples in distributed. I'd prefer to make this change in a separate PR.

PR #808, opened 2020-08-01
Add fizz/buzz example.

i25ffz:
Add fizz/buzz example, you can use --num-digits to assign the training/testing digit.


soumith:
thanks for the contribution, but it doesn't seem like a good representative example of anything people would want to do, and hence learn from

PR #809, opened 2020-08-01
Distributed example on C++ API (LibTorch) 

soumyadipghosh:
This PR adds an example involving distributed training using MPI on the C++ frontend similar to DistributedDataParallel in Python. This topic was raised in the forums [here](https://discuss.pytorch.org/t/libtorch-c-mpi-example/39671). Right now, this code is CPU only. 

Please let me know if this PR can be a worthwhile contribution.

cc @yf225 @pietern


yf225:
@glaringlee @mrshenli Curious would this be a good addition? Thanks!

glaringlee:
@yf225 This is good to me. I saw there is only 1 check (Run Examples) within this repo, is that as designed?
@mrshenli can you check the logic here, it seems fine to me though.

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+          //Reset gradients
+          model->zero_grad();
+
+          //Execute forward pass
+          auto prediction = model->forward(ip);
+
+          auto loss = torch::nll_loss(torch::log_softmax(prediction, 1) , op);
+
+          /*
```
mrshenli, line 116: debug vestige?

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,201 @@
...
+          */
+
+          //Backpropagation
+          loss.backward();
+
+          //Averaging the gradients of the parameters in all the processors
+          for (auto &param : model->named_parameters())
+          {
+              MPI_Allreduce(MPI_IN_PLACE, param.value().grad().data_ptr(),
+               param.value().grad().numel(),
```
mrshenli, line 130: could you please use clang-format to format code?

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,201 @@
...
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+  //end MPI variables
+
+  // Timer variables
+  auto tstart = 0.0;
+  auto tend = 0.0;
+
+  //TRAINING
+  //Read train dataset
+  std::string filename = "/afs/crc.nd.edu/user/s/sghosh2/Public/ML/mnist/data";
```
mrshenli, line 63: Should we make this an input arg instead of hard-coding a path?

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,201 @@
...
+          }
+          */
+
+          //Backpropagation
+          loss.backward();
+
+          //Averaging the gradients of the parameters in all the processors
+          for (auto &param : model->named_parameters())
+          {
+              MPI_Allreduce(MPI_IN_PLACE, param.value().grad().data_ptr(),
```
mrshenli, line 129: cc @agolynski, have we exposed ProcessGroup C++ APIs? Can users call allreduce directly from process group instances?

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,201 @@
...
+               param.value().grad().numel(),
+                mpiDatatype.at(param.value().grad().scalar_type()),
+                MPI_SUM, MPI_COMM_WORLD);
```
mrshenli, line 132: Curious, is this MPI op blocking? Will it help speed up training if we launch all asynchronously and then wait for all?

soumyadipghosh: This MPI op is blocking. If we want to do it asynchronously, we can use [MPI_Iallreduce](https://www.mpich.org/static/docs/v3.1/www3/MPI_Iallreduce.html). However, that would just overlap the computation with communication. We still need to use [MPI_Wait](https://www.mpich.org/static/docs/v3.1/www3/MPI_Wait.html) to ensure that the non-blocking operation has completed before training moves to the next epoch.

cpp/distributed/dist-mnist.cpp
soumyadipghosh: Forgot to remove that! Should I put the path as a user input or search for it in a subfolder as shown in this [example](https://github.com/pytorch/examples/blob/master/cpp/mnist/mnist.cpp) ?

glaringlee:
@yf225 @mrshenli 
I see there is a CI like test in this repo which run all the examples, any change that fail the execution will be detected in the test. For silence failure (logic change but won't stop the execution), can we add assert within the example so it breaks if any bc breaking change is made?

cpp/distributed/dist-mnist.cpp
soumyadipghosh: yeah, removed now.

cpp/distributed/dist-mnist.cpp
soumyadipghosh: done!

mrshenli:
Do we need to add this new test to some `.sh` file, or will the CI automatically detect and include all new tests?

glaringlee:
Should add dist-mnist after this line I think and do proper cleanup:
https://github.com/pytorch/examples/blob/master/run_python_examples.sh#L189
cc @yf225 @mrshenli @soumyadipghosh

soumyadipghosh:
Also, for running tests, an MPI installation is necessary. How will that be added to the CI pipeline?

glaringlee:
I think the CI is totally controlled by run_python_examples.sh. MPI can be installed when installing dependencies.
cc @seemethere

soumyadipghosh:
cc @yf225 @glaringlee @mrshenli Can someone help me with the next steps ? I am not sure how to add the dependencies and the tests.

glaringlee:
@malfet Are you familiar with this repo? Should we update run_python_examples.sh (add this test in) to make this test running on each code update? cc @mrshenli

soumyadipghosh:
Can someone please help me with the next steps? I think the [run_python_examples.sh](https://github.com/pytorch/examples/blob/master/run_python_examples.sh) is only relevant for the python examples. It would be better to create something like run_cpp_examples.sh to test the cpp examples including this one.
cc @yf225 @glaringlee @mrshenli @malfet @seemethere @soumith

cpp/distributed/dist-mnist.cpp
soumyadipghosh: changed to "../data"

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,192 @@
...
+      auto prediction = model->forward(ip);
+
+      auto loss = torch::nll_loss(torch::log_softmax(prediction, 1), op);
+
+      // Backpropagation
+      loss.backward();
+
+      // Averaging the gradients of the parameters in all the processors
+      for (auto& param : model->named_parameters()) {
+        MPI_Allreduce(
```
osalpekar, line 119: Can we create an MPI Process Group and call the process group `all_reduce` function instead?

cpp/distributed/dist-mnist.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+      // Execute forward pass
+      auto prediction = model->forward(ip);
+
+      auto loss = torch::nll_loss(torch::log_softmax(prediction, 1), op);
+
+      // Backpropagation
+      loss.backward();
+
+      // Averaging the gradients of the parameters in all the processors
```
osalpekar, line 117: I'm wondering whether we should add a comment here indicating that while this approach will allow synchronous data parallel training, it will lag behind DDP in performance since it synchronizes parameters after the backward pass whereas DDP overlaps synchronizing parameters and computing gradients in the backwards pass.

osalpekar (commented):
Thanks for giving an update on your work in the PyTorch forums! Adding some thoughts that might improve this example

cpp/distributed/dist-mnist.cpp
soumyadipghosh: @osalpekar I can add that comment. Also I can call MPI_Iallreduce before the backward pass which is a non-blocking call, then call MPI_Wait for that call after the backward pass. This will ensure that backward pass and synchronizing parameters are overlapped. What do you say?

cpp/distributed/dist-mnist.cpp
soumyadipghosh: @osalpekar We can do that but will that help in some way?

cpp/distributed/dist-mnist.cpp
osalpekar: I don't think that will have the same effect here. `loss.backward` is what actually computes the gradients so I don't think we can launch the synchronization collective before we've even started computing gradients.

DDP has a notion of buckets. We've instrumented the backward pass in DDP such that, once a bucket has been filled with gradients, we launch the allreduce. While the allreduce call is executing, we continue computing the next gradients to fill the next bucket, and so on. Trying to re-create this behavior efficiently is probably beyond the scope of this example so we can just stick to a blocking allreduce after the backward pass, but just in case somebody tries to measure the performance, we should include the comment so that they know this approach will be slower than DDP.

cpp/distributed/dist-mnist.cpp
osalpekar: @soumyadipghosh good question - most of the reasons below exist for Python - let's check with @agolynski about ProcessGroup C++ API just to make sure this is allowed.

Using ProcessGroup is the preferred way of launching collectives and communication operations in PyTorch distributed. This way you can create a ProcessGroup using any underlying communication library you prefer (MPI, NCCL, Gloo, etc.), and simply call any of the collectives (allreduce, broadcast, etc.). Not only are you not tied to any comm library at the user program level, but you also gain some other benefits depending on the specific ProcessGroup type being used (error handling, timeout checking, perf optimizations, planned support for async Futures, etc.).

cpp/distributed/dist-mnist.cpp
soumyadipghosh: I see! Added the comment :)

cpp/distributed/dist-mnist.cpp
soumyadipghosh: @osalpekar Any update on how to proceed about the ProcessGroup C++ API?

soumyadipghosh:
@osalpekar @mrshenli Since we haven't heard back about the ProcessGroup C++ APIs (see comments above), I created a separate file that involves NCCL for GPU communication that is somewhat based on the NCCL Example 2 [here](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/examples.html). I am not sure how to combine the CPU and GPU versions into one file because the NCCL routines will lead to errors when run on CPUs. Any ideas on the next steps?

cpp/distributed/dist-mnist.cpp
osalpekar: @soumyadipghosh Apologies for the delay! [Here's](https://github.com/pytorch/pytorch/blob/c115957df086fd795abaae18c6b9bc1e80928a18/torch/lib/c10d/ProcessGroupMPI.hpp) a link to the ProcessGroupMPI header file. You can simply construct the ProcessGroup and then call any collectives on that process group (which are implemented as class methods - for example, allreduce). If the MPI you are using is CUDA-aware, you can use CUDA tensors.

We also have a [NCCL ProcessGroup](https://github.com/pytorch/pytorch/blob/c115957df086fd795abaae18c6b9bc1e80928a18/torch/lib/c10d/ProcessGroupNCCL.hpp) that you can as above for using the NCCL backend. This way you will not have to call into the NCCL functions for initialization, creating communicators, etc. - it is all done for you, so you can simply create the ProcessGroup and then use the provided collectives (for example, `ProcessGroupNCCL::allreduce`). 

If you want to use both CPUs and GPUs while training, the best option would be Gloo (we also have a ProcessGroupGloo in the same folder as the above 2) and another option is to use MPI if your version of MPI is CUDA-aware.

osalpekar:
> @osalpekar @mrshenli Since we haven't heard back about the ProcessGroup C++ APIs (see comments above), I created a separate file that involves NCCL for GPU communication that is somewhat based on the NCCL Example 2 [here](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/examples.html). I am not sure how to combine the CPU and GPU versions into one file because the NCCL routines will lead to errors when run on CPUs. Any ideas on the next steps?

I added a comment above about the ProcessGroup C++ APIs - I think these should significantly simplify some of the initialization boilerplate that's here for NCCL. For the sake of this example, adding support for multiple ProcessGroups might be a bit too much since these examples should demonstrate to users how to efficiently use PyTorch Distributed training - we can stick to MPI for this example and include a comment that NCCL or Gloo can be used as alternatives. I think we should be in good shape to merge this in once we move to the MPI ProcessGroup.

We also have an ongoing discussion about native DDP C++ support (https://github.com/pytorch/pytorch/issues/48959), so please feel free to contribute ideas there as well!

cpp/distributed/dist-mnist.cpp
```diff
@@ -1,19 +1,11 @@
+#include <c10d/ProcessGroupMPI.hpp>
```
soumyadipghosh, line 1: this line leads to error!

cpp/distributed/dist-mnist.cpp
```diff
@@ -45,16 +37,17 @@ struct Model : torch::nn::Module {
...
-  int rank, numranks;
-  MPI_Init(&argc, &argv);
-  MPI_Comm_size(MPI_COMM_WORLD, &numranks);
-  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-  // end MPI variables
+  initMPIOnce();
 
+  int rank = atoi(getenv("RANK"));
+  int numranks = atoi(getenv("SIZE"));
+  c10d::ProcessGroupMPI pg(rank, numranks, MPI_COMM_WORLD);
```
soumyadipghosh, line 44: Not sure if the above is suitable to initialize the MPI environment and create the ProcessGroup API

cpp/distributed/dist-mnist.cpp
```diff
@@ -89,8 +82,10 @@ int main(int argc, char* argv[]) {
   // Number of epochs
   size_t num_epochs = 10;
 
+  /*
   // start timer
   tstart = MPI_Wtime();
+  */
```
soumyadipghosh, line 88: Since we are not using MPI directly, how to call an MPI primitive such as MPI_Wtime() separately in the code to measure the training time?

cpp/distributed/dist-mnist.cpp
```diff
@@ -120,14 +115,7 @@ int main(int argc, char* argv[]) {
...
       for (auto& param : model->named_parameters()) {
-        MPI_Allreduce(
-            MPI_IN_PLACE,
-            param.value().grad().data_ptr(),
-            param.value().grad().numel(),
-            mpiDatatype.at(param.value().grad().scalar_type()),
-            MPI_SUM,
-            MPI_COMM_WORLD);
-
+        pg.allreduce(param.value().grad());
```
soumyadipghosh, line 118: How to add the AllreduceOptions for the SUM operator?

cpp/distributed/dist-mnist.cpp
```diff
@@ -192,5 +182,5 @@ int main(int argc, char* argv[]) {
               << std::endl;
   } // end rank 0
 
-  MPI_Finalize();
+  mpiExit();
```
soumyadipghosh, line 185: Not sure if this is the right way to finalize the MPI environment?

soumyadipghosh:
@osalpekar I agree that we can just stick to MPI for this example and so I removed the NCCL code for now. Regarding the move to ProcessGroupMPI, I made some initial changes to the code based on the c10d example [here](https://github.com/pytorch/pytorch/blob/master/torch/lib/c10d/example/allreduce.cpp) but it doesn't compile! I have a lot of questions which I put as comments in the code, can you please clarify them?

cpp/distributed/dist-mnist.cpp
lasagnaphil: Maybe is it related to the fact that the default PyTorch installation excludes support for MPI? It says that you need to recompile PyTorch from scratch if you want to use the MPI backend for torch.distributed. (https://pytorch.org/tutorials/intermediate/dist_tuto.html#communication-backends) As I'm also having this exact same issue, I'll try to do that tomorrow.

cpp/distributed/dist-mnist.cpp
osalpekar: I think `ProcessGroupMPI::createProcessGroupMPI` would probably encompass all the initialization. It calls `initMPIOnce` for you and returns a correctly instantiated ProcessGroup

cpp/distributed/dist-mnist.cpp
osalpekar: Since this is an illustrative example for future users, let's leave out timing code. For your personal use, you can still call `MPI_Wtime()` before and after whichever function you want to measure the time of right? For example, before and after `pg.allreduce`?

cpp/distributed/dist-mnist.cpp
osalpekar: `AllreduceOptions` is a second optional argument for the `allreduce` function.
You can create the AllreduceOptions struct and set the `reduceOp = ReduceOp::SUM`, but this is already the default so you should be able to leave this as is.

cpp/distributed/dist-mnist.cpp
osalpekar: `mpiExit` is called for you, you don't need to call it separately.

cpp/distributed/dist-mnist.cpp
soumyadipghosh: Ok done.

cpp/distributed/dist-mnist.cpp
soumyadipghosh: I was measuring the time for training. But for the sake of simplicity, I removed it.

cpp/distributed/dist-mnist.cpp
soumyadipghosh: Alright. Also adding the waiting mechanism as stated in [https://github.com/pytorch/pytorch/blob/master/torch/lib/c10d/test/ProcessGroupMPITest.cpp](https://github.com/pytorch/pytorch/blob/master/torch/lib/c10d/test/ProcessGroupMPITest.cpp)

cpp/distributed/dist-mnist.cpp
soumyadipghosh: Removed it

cpp/distributed/dist-mnist.cpp
soumyadipghosh: @osalpekar How can this be fixed? I am not sure if @lasagnaphil ' s idea will work since Libtorch doesn't need to be compiled?

lasagnaphil: The libtorch installation files that pytorch.org distributes are all precompiled libraries (with their accompanying headers). Chances are that even if you use the system's package manager (such as the Arch Linux AUR), they will use those official precompiled libraries. And from what I've seen, these do not include any support for MPI (for example, it's missing the header ProcessGroupMPI.hpp)

It seems that the build servers for pytorch.org do not include MPI support, which is understandable because there are multiple MPI implementations (OpenMPI, MPICH, MVAPICH, ...) and it's impractical to build and distribute the library separately for each library. So the only option would be to build and install libtorch from scratch. Also, the CMake script for c10d seems to include MPI support for the library only if a valid MPI installation is found (https://github.com/pytorch/pytorch/blob/c215ffb6a2de1c85a62a146a9e282908335ba3ed/torch/lib/c10d/CMakeLists.txt#L146), so you need to make sure an MPI installation exists on your system before calling ``python setup.py install``.

beniz: As a side note, this might indicate that NCCL would make for a simpler example, though certainly less interesting in terms of showing how to deal with some more internals (e.g. intialization, ...).

soumyadipghosh: @lasagnaphil I see, thanks! Were you able to compile libtorch from scratch using any MPI?

lasagnaphil: Compiling works now, as the headers (at least) are now properly installed on the system. However, I'm still having a linking error (``undefined reference to `c10d::ProcessGroupMPI::createProcessGroupMPI``) that I need to investigate. I tried to link the ``c10d`` CMake target to my executable but it says that it doesn't exist, which is a mystery.

lasagnaphil: Ah, I figured it out. It seems like the file ``libc10d.a`` wasn't copied properly into the installation folder. Also, it seems that the particular static library needs to be linked directly with the exact path as the following

```
target_link_libraries(dist-mnist ${CMAKE_PREFIX_PATH}/lib/libc10d.a)
```
After this, and fixing a couple of more errors, I was able to successfully run the example (training seems to be doing all fine). Here's the pull request for the fix: https://github.com/soumyadipghosh/examples/pull/1

soumyadipghosh: @lasagnaphil Great! Can you share how you compiled libtorch from scratch using MPI?

lasagnaphil: It took a while to figure out how to compile libtorch cleanly, I'll provide you a shell script: https://gist.github.com/lasagnaphil/3e0099816837318e8e8bcab7edcfd5d9

There are a lot of flags, and you might need to tweak some of them depending on your PC / usecase.

Take in mind that it takes quite some CPU time to compile! (For my Ryzen 7 2700X system, it took about 20~30 minutes)

osalpekar: @soumyadipghosh The PR should be in good shape after @lasagnaphil's patch - we should be good to go ahead and approve. Are you still seeing build issues?

soumyadipghosh: @osalpekar Sorry, I didn't have the chance to verify - will do it soon. For now, I have merged @lasagnaphil 's patch and added the link to the shell script for build in the Readme. We should be good to go ahead now. If I have build issues later, I will create an issue.

osalpekar: @soumyadipghosh No problem, I'm accepting this for now. Would be great if we could verify the build before merging

osalpekar (approved):
Looks great - thanks for your work on this example @soumyadipghosh!

soumyadipghosh:
@osalpekar Thanks for all your help! I have verified the build and everything works great! Ready for merging :)

soumyadipghosh:
@osalpekar So how does this get merged?

PR #811, opened 2020-08-05
Run CI daily

mattip:
xref pytorch/pytorch#32004

Run github actions CI once a day (at 3:00) and if it fails, open an issue. mattip/pytorch-examples#21 is a forced example of such an issue. A few things to note:
- The main purpose of this action is to consistently test this repo against pytorch nightly builds, to see if some new pytorch "feature" breaks one of the examples.
- Anyone who forks this repo will, by default, get this scheduled action. I added a note to the README. We could also add a step to emit an issue `if {{ github.repository_owner != 'pytorch' }}` which would hopefully get people's attention :). By default forks do not have issues, so I am not sure what would happen. Forking this repo does seem like a popular thing to do: there are 64.k forks. Searching for a way to disable via the yml led me to [this discussion](https://github.community/t/how-can-i-disable-a-github-action/17049) which seems to indicate it cannot be done right now. Another option would be to move from github actions (which are available on all public repos) to travis or circleci, which would require action on part of the fork to enable it. I can pivot this PR to do that, but a admin would have to enable another CI service.
- I _hope_ someone is getting mails on all new issues in this repo, otherwise the issue will not get much attention.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,10 +1,14 @@
 name: Run Examples
 
 on:
- push:
-   branches: [ master ]
- pull_request:
-   branches: [ master ]
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    # Every day at 3:00am
+    - cron: '0 3 * * *'
+        
 
 jobs:
   test:
@@ -27,5 +31,12 @@ jobs:
     - name: Run Tests
       run: |
         ./run_python_examples.sh "install_deps,run_all,clean"
-
+    - name: Open issue on failure
+      if: ${{ failure() && github.event_name  == 'schedule' }}
+      uses: rishabhgupta/git-action-issue@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        title: Daily CI failed
+        body:  Commit ${{ github.sha }} daily scheduled [CI run](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}) failed, please check why
+        assignees: ''
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # PyTorch Examples
 ![Run Examples](https://github.com/pytorch/examples/workflows/Run%20Examples/badge.svg)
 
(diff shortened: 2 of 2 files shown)


mattip:
ping @brianjo

rgommers:
> Another option would be to move from github actions (which are available on all public repos) to travis or circleci, which would require action on part of the fork to enable it. I can pivot this PR to do that, but a admin would have to enable another CI service.

That's going to generate a lot of noise; with 6.5k forks and a good fraction of new forks missing that README note, that's a lot of cron jobs being kicked off. 

Can't you use a conditional with, e.g., `github.repository_owner` (see https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context)

mattip:
> Can't you use a conditional ...

I don't think a [conditional](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) on a [schedule](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule) directive works, so we would have to add one to each step. And even then the cron job would trigger, it would start a worker and check out the repo, and only then try each step to discover they all are skipped.

Perhaps running the cron job on another CI system would be a better solution.

brianjo:
@mattip I think you could add  `if: github.repository == ‘pytorch/examples’ ` to the script after the schedule directive triggers it and it will just skip the rest of the script on a fork.

mattip:
@brianjo what about [setting up azure pipelines](https://github.com/marketplace/azure-pipelines) for this repo, then only forks that turn on azure pipelines would run the cron job in the first place. Its features are very comparable to github actions, in fact they share the same infrastrucure.

brianjo:
@mattip This popped up yesterday on a fork I did. I think we might not have to worry about it. 

image

mattip:
Ahh, cool! It seems someone thought about this problem before me :)

So in that case I think this is good to go

soumith:
thanks @mattip !

mattip:
I guess we can consider this done when a green checkmark shows up on a daily run in the [actions tab](https://github.com/pytorch/examples/actions)

mattip:
... and when it fails an issue is properly opened.

PR #820, opened 2020-09-02
use reshape instead of view in accuracy

ngimel:
Per title. Layout of `pred` is now correctly propagated so that `correct` is not contiguous.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -420,7 +420,7 @@ def accuracy(output, target, topk=(1,)):
 
         res = []
         for k in topk:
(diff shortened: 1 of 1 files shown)


imagenet/main.py
```diff
@@ -420,7 +420,7 @@ def accuracy(output, target, topk=(1,)):
 
         res = []
         for k in topk:
-            correct_k = correct[:k].view(-1).float().sum(0, keepdim=True)
+            correct_k = correct[:k].reshape(-1).float().sum(0, keepdim=True)
```
gchanan, line 423: wouldn't something like `correct[:k].float().sum().unsqueeze(0)` be nicer?

PR #823, opened 2020-09-21
Fix RPC examples to use TensorPipeRpcBackendOptions.

pritamdamania87:
Summary: Since TensorPipe is going to be the default backend in 1.7, we
should change our examples to use this backend. Some examples were
failing since the default was TensorPipe but we were passing in
ProcessGroupBackendOptions.

Test Plan: Run the test examples.


lw (approved):
And thanks for doing these ones as well!

PR #825, opened 2020-09-25
Fix random seed bug of DDP in the ImageNet example

LeeDoYup:
the random seed has to be set in the main_worker, not in the `def main()`.
I found that although the seed is set in `def main()`,
each process in distributed training has individual & different seeds.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -81,9 +81,6 @@ def main():
     args = parser.parse_args()
 
     if args.seed is not None:
-        random.seed(args.seed)
-        torch.manual_seed(args.seed)
-        cudnn.deterministic = True
         warnings.warn('You have chosen to seed training. '
                       'This will turn on the CUDNN deterministic setting, '
                       'which can slow down your training considerably! '
@@ -115,6 +112,11 @@ def main():
 def main_worker(gpu, ngpus_per_node, args):
     global best_acc1
     args.gpu = gpu
+    
+    if args.seed is not None:
+        random.seed(args.seed)
+        torch.manual_seed(args.seed)
+        cudnn.deterministic = True
 
     if args.gpu is not None:
         print("Use GPU: {} for training".format(args.gpu))


sammiie:
To fix the random seed bug of DDP (Distributed Data Parallel) in the ImageNet example, you'll need to make changes to the code. I'll provide a code solution below that addresses this issue

Identify the Issue:
First, you need to identify the specific issue related to the random seed bug. Look for the code section or function where random seeds are set or used incorrectly.

Import Necessary Libraries:
Import the necessary libraries and modules for setting random seeds.

Set Random Seed:
Set the random seed at the beginning of your script or in the appropriate location. This ensures that the random number generator produces reproducible results across different runs. For PyTorch, you can use the following code:

import torch
import random
import numpy as np

# Set random seeds for PyTorch, Python, and NumPy
torch.manual_seed(0)
random.seed(0)
np.random.seed(0)


Make sure to use the same seed value (e.g., 0) for all three libraries.

Check for DDP Initialization:
If DDP initialization is causing the issue, you may need to wrap it with a condition to only enable DDP when it's appropriate. For example:

if __name__ == "__main__":
    # Set random seeds
    torch.manual_seed(0)
    random.seed(0)
    np.random.seed(0)

    # Your code to load the model and data

    # Check if DistributedDataParallel (DDP) is needed
    if use_ddp:
        # Initialize DDP here
        model = torch.nn.parallel.DistributedDataParallel(model)
    
    # Rest of your code for training or evaluation



To fix the random seed bug of DDP (Distributed Data Parallel) in the ImageNet example, you'll need to make changes to the code. I'll provide a code solution below that addresses this issue:

Identify the Issue:
First, you need to identify the specific issue related to the random seed bug. Look for the code section or function where random seeds are set or used incorrectly.

Import Necessary Libraries:
Import the necessary libraries and modules for setting random seeds.

Set Random Seed:
Set the random seed at the beginning of your script or in the appropriate location. This ensures that the random number generator produces reproducible results across different runs. For PyTorch, you can use the following code:

python
Copy code
import torch
import random
import numpy as np

# Set random seeds for PyTorch, Python, and NumPy
torch.manual_seed(0)
random.seed(0)
np.random.seed(0)
Make sure to use the same seed value (e.g., 0) for all three libraries.

Check for DDP Initialization:
If DDP initialization is causing the issue, you may need to wrap it with a condition to only enable DDP when it's appropriate. For example:
python
Copy code
if __name__ == "__main__":
    # Set random seeds
    torch.manual_seed(0)
    random.seed(0)
    np.random.seed(0)

    # Your code to load the model and data

    # Check if DistributedDataParallel (DDP) is needed
    if use_ddp:
        # Initialize DDP here
        model = torch.nn.parallel.DistributedDataParallel(model)
    
    # Rest of your code for training or evaluation
Testing:
After making these changes, test your code thoroughly to ensure that the random seed bug is fixed and that your script produces reproducible results.

Commit Changes:
If you are working with version control (e.g., Git), commit your changes with a clear message describing the bug fix.

These steps should help you fix the random seed bug in the DDP implementation for the ImageNet example. Be sure to adapt the code to the specific structure and requirements of your project.

PR #832, opened 2020-10-10
Fix unused --test-batch-size command line argument

### LINKED ISSUE #157: two lines of code in mnist/main.py
There are two arguments called batch_size and test_batch_size:
`parser.add_argument('--batch-size', type=int, default=64, metavar='N',
                    help='input batch size for training (default: 64)')`
`parser.add_argument('--test-batch-size', type=int, default=1000, metavar='N',
                    help='input batch size for testing (default: 1000)')`
but batch_size is used here:
`test_loader = torch.utils.data.DataLoader(
    datasets.MNIST('../data', train=False, transform=transforms.Compose([
                       transforms.ToTensor(),
                       transforms.Normalize((0.1307,), (0.3081,))
                   ])),
    batch_size=args.batch_size, shuffle=True, **kwargs)`

Also, what does this line(line 105) do:
`test_loss = test_loss`

and it seems that `epoch` is not used in test().

pbelevich:
Fixes #157

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -100,12 +100,14 @@ def main():
 
     device = torch.device("cuda" if use_cuda else "cpu")
 
-    kwargs = {'batch_size': args.batch_size}
+    train_kwargs = {'batch_size': args.batch_size}
+    test_kwargs = {'batch_size': args.test_batch_size}
     if use_cuda:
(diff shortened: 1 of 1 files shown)


soumith:
thanks!

PR #658, opened 2019-11-05
Improving mnist example model

fshabashev:
I look on Keras examples and found that they have a better performing model for Mnist. 
https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py

I reimplemented their approach in PyTorch and got a better performing model.

I compared the new model with the current pytorch example.
Here are the test results:
```
Pytorch original mnist example:
Test set: Average loss: 0.0315, Accuracy: 9895/10000 (99%)
Test set: Average loss: 0.0310, Accuracy: 9894/10000 (99%)
Test set: Average loss: 0.0315, Accuracy: 9896/10000 (99%)
Test set: Average loss: 0.0317, Accuracy: 9891/10000 (99%)
Test set: Average loss: 0.0314, Accuracy: 9894/10000 (99%)
Test set: Average loss: 0.0324, Accuracy: 9894/10000 (99%)
Test set: Average loss: 0.0317, Accuracy: 9892/10000 (99%)
Test set: Average loss: 0.0314, Accuracy: 9895/10000 (99%)
Test set: Average loss: 0.0310, Accuracy: 9895/10000 (99%)
Test set: Average loss: 0.0315, Accuracy: 9893/10000 (99%)
Test set: Average loss: 0.0319, Accuracy: 9894/10000 (99%)

Pytorch proposed model example:
Test set: Average loss: 0.0261, Accuracy: 9924/10000 (99%)
Test set: Average loss: 0.0276, Accuracy: 9914/10000 (99%)
Test set: Average loss: 0.0279, Accuracy: 9921/10000 (99%)
Test set: Average loss: 0.0276, Accuracy: 9912/10000 (99%)
Test set: Average loss: 0.0279, Accuracy: 9918/10000 (99%)
Test set: Average loss: 0.0272, Accuracy: 9920/10000 (99%)
Test set: Average loss: 0.0273, Accuracy: 9917/10000 (99%)
Test set: Average loss: 0.0270, Accuracy: 9918/10000 (99%)
Test set: Average loss: 0.0264, Accuracy: 9919/10000 (99%)
Test set: Average loss: 0.0275, Accuracy: 9919/10000 (99%)
```
As you can see, accuracy for new model is higher, while logloss is lower.
I'm suggest to replace the current version with the better performing model. 
The new model is not much more complex and training time is somewhat similar. 
What do you think?

PR #833, opened 2020-10-10
MNIST example: use regular dropout rather than dropout2d

eli-osherovich:
Using dropout2d does not really make sense for the second dropout since the data is flattened at that point (but it works accidentally). Furthermore, since this example copies the one from Keras (https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py) it might be better to follow that example (regular dropout). 

No noticeable performance changes.


soumith:
the example started as a migration from torch7, but at some point, someone must've contributed a patch to mimic the Keras model.

soumith:
looks like https://github.com/pytorch/examples/pull/658 was where we moved over to the Keras-example style model

PR #841, opened 2020-11-11
fixed conditions about using single GPU

sousquared:
1. A better condition for single GPU (when your machine has only one GPU)
I found that even if your machine has only one GPU and you don't specify GPU id (e.g. `--gpu 0`), model is passed to `DataParallel`.
So I added a condition so that your model is not parallel even when you don't use `--gpu` option.
```python
elif torch.cuda.device_count() == 1:
        model = model.cuda()
```
2. Removing redundant condition
I fixed conditions about putting torchTensors on CUDA memory. We don't need `if args.gpu is not None` here:
```python 
if args.gpu is not None:
    images = images.cuda(args.gpu, non_blocking=True)
```
According [pytorch document](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.cuda), in `torchTensors.cuda(device=None)` is also acceptable.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -159,6 +159,8 @@ def main_worker(gpu, ngpus_per_node, args):
     elif args.gpu is not None:
         torch.cuda.set_device(args.gpu)
         model = model.cuda(args.gpu)
+    elif torch.cuda.device_count() == 1:
+        model = model.cuda()
     else:
         # DataParallel will divide and allocate batch_size to all available GPUs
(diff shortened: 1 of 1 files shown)


cbalioglu:
@msaroufim ideally for each DP example we should create a counterpart example that uses DDP. We haven't deprecated DP yet, since we don't have enough resources right now to work on feature parity. However we already encourage our users to migrate to DDP if they can. We would be sending a mixed signal if we had examples covered by DP, but not DDP.

PR #842, opened 2020-11-17
[WIP] Fix rpc param server example

rohan-varma:
Open 3 separate terminals and run:
python rpc_parameter_server.py --world_size=3 --rank=0
python rpc_parameter_server.py --world_size=3 --rank=1
python rpc_parameter_server.py --world_size=3 --rank=2

The main issue seems to be that one trainer calls optimizer.step() and updates the weights before the other trainer calls backward, which results in the autograd error. 

This fix makes it so that only one trainer operates on the same model one at a time. I.e. only one trainer does all of fwd, backward, and opt, and only then is the next trainer allowed to do its fwd/backward/opt step. This is done hackily by the trainers waiting for each other with cond vars.

Note: This fix is just for discussion, not intended to be merged.

### DIFF
--- a/distributed/rpc/parameter_server/rpc_parameter_server.py
+++ b/distributed/rpc/parameter_server/rpc_parameter_server.py
@@ -19,7 +19,10 @@ from torchvision import datasets, transforms
 class Net(nn.Module):
     def __init__(self, num_gpus=0):
         super(Net, self).__init__()
-        print(f"Using {num_gpus} GPUs to train")
+        if num_gpus > 0:
+            print(f"Using {num_gpus} to train")
+        else:
+            print("CPU-only training")
         self.num_gpus = num_gpus
         device = torch.device(
             "cuda:0" if torch.cuda.is_available() and self.num_gpus > 0 else "cpu")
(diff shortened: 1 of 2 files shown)


Weigaa:
Thank you for solving this problem, but does this solution mean that multiple workers train in series and cannot use multiple nodes to speed up the training process in parallel? e.g. I have 3 CPUs,1 PS, and 2  workers, and the workers' speed is v1 and v2, total training data size = n,now the training time = n/2v1 + n/2v2 but what we expect training time = n/2min(v1,v2)?

Arunpoochelvan (changes requested):
Change line number 201 
    device = torch.device("cuda:0" if model.num_gpus > 0 
    device = torch.device("cuda:0" if model.num_gpus > 0
        and torch.cuda.is_available() else "cpu")

to 

    device = torch.device("cuda:0" if model.num_gpus > 0 
        and torch.cuda.is_available() else "cpu")

rohan-varma:
> Change line number 201
> device = torch.device("cuda:0" if model.num_gpus > 0
> device = torch.device("cuda:0" if model.num_gpus > 0
> and torch.cuda.is_available() else "cpu")
> 
> to
> 
> ```
> device = torch.device("cuda:0" if model.num_gpus > 0 
>     and torch.cuda.is_available() else "cpu")
> ```

Sorry for the late reply, but not sure if I follow this. Aren't these the same lines of code?

msaroufim:
Hi @rohan-varma the RPC example we have in this repo is from 1.4 do you think we should just go ahead and remove it?

msaroufim:
Closing since this is now stale

PR #849, opened 2020-11-30
Replaced After I replaced Line 64 in generate.py word_idx = torch.multinomial(word_weights, 1)[0] with word_idx = torch.argmax(word_weights).item()

Koushikk-24:
word_idx=torch.argmx(word_weights.item()


soumith:
this change doesn't make sense. it maximizes the generation always, instead of keeping a randomness to it.

PR #859, opened 2020-12-21
Use PyTorch Step Scheduler

developer0hye:
Use PyTorch Step Scheduler


msaroufim (changes requested):
I like this change, do you mind just fixing the import statements and not getting `*` in on line 14 and could you confirm that the test still runs

developer0hye:
@msaroufim 
Thanks for your review! I modified it.

msaroufim (approved):

PR #864, opened 2020-12-30
fast neural style: run onnxmodel using onnxruntime

kleinicke:
The old onnx implementations seemed to be outdated and didn't work properly for me.
I set the model to evaluation mode before transforming to onnx, I set the onnx opset_version to 11 (below there was a warning).
By running the onnx model I got the warning, that onnx_caffe2.backend only supports opset_version 2.
Therefore I followed the tutorial: https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html to run the onnx model with onnxruntime instead.

I wasn't sure if I should leave the old function in the file with an option to run it, but I deleted to keep the file clean.

The new implementation only performs the final style transfer using the onnx model on the cpu, but that's the purpose of real time style transfer anyway.

### DIFF
--- a/fast_neural_style/neural_style/neural_style.py
+++ b/fast_neural_style/neural_style/neural_style.py
@@ -131,7 +131,7 @@ def stylize(args):
     content_image = content_image.unsqueeze(0).to(device)
 
     if args.model.endswith(".onnx"):
-        output = stylize_onnx_caffe2(content_image, args)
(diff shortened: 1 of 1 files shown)


kleinicke:
I can also reformulate what I've changed.
I've replaced the outdated onnx_caffe2.backend with the current onnxruntime, since the old implementation doesn't seem to work anymore.

PR #875, opened 2021-01-24
Fix distributed/rpc/rnn unused import, typos, formatting

pbelevich:


mrshenli (approved):
Thanks for fixing and improving!!

PR #876, opened 2021-01-25
Fix distributed/rpc/pipeline unused import, typos, formatting

pbelevich:


distributed/rpc/pipeline/main.py
```diff
@@ -32,9 +31,17 @@ def conv1x1(in_planes, out_planes, stride=1):
     """1x1 convolution"""
     return nn.Conv2d(in_planes, out_planes, kernel_size=1, stride=stride, bias=False)
 
+
+def get_cuda_if_available(i):
+    assert i >= 0
+    if torch.cuda.is_available():
+        return f"cuda:{min(i, torch.cuda.device_count() - 1)}"
```
mrshenli, line 38: this is not the intention. This example requires at least two GPUs to run so that we can show the benefits of RPC-powered pipeline parallel across devices. If that's confusing, maybe we can add this requirement to the readme?

pbelevich: I totally understand the intension, that why this example will work as before for those who have 2 GPUs, but will also be runnable on the environments with less then 2 GPUs (most of the ordinary users and FB Sandcastle)

msaroufim:
Hi @pbelevich would you still like to contribute this example or should we close it? Our RPC example is still quite cold, dates back to pytorch 1.4 it looks like

PR #881, opened 2021-02-02
[FX] Start fx folder and add native interpreter example

jamesr66a:
Creates an `fx` folder and adds an example for lowering PyTorch code into a native runtime

### DIFF
--- a/fx/native_interpreter/README.md
+++ b/fx/native_interpreter/README.md
@@ -1,0 +1,35 @@
+# Converting PyTorch Code to a Native Runtime With FX and TorchScript Custom Classes
+
+In this example, we are going to build a pipeline that does the following things:
+
+1. Converts (or “lowers”) code in a PyTorch module into another representation (we will define the representation within the example)
+2. Registers an interpreter for that code representation that can be used in TorchScript or Python
+3. Wrap the converted code into a format that can still be used in TorchScript compilation.
+
+We are going to build up a trivial interpreter for this example, but you can imagine extending the same process to work with more sophisticated backends, ones which may do code optimization or offloading to an accelerator.
+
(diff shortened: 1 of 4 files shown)


fx/native_interpreter/use_interpreter.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # ===== Stage 3: Create a wrapper GraphModule around the interpreter =====
+    class WrapperModule(torch.nn.Module):
+        def __init__(self, interpreter):
+            super().__init__()
+            self.interpreter = interpreter
+
+    wrapper = WrapperModule(interpreter)
+
+    # Create a graph that: 1) Takes function arguments 2) Invokes the interpreter
```
zdevito, line 89: probably add a comment that the reason to do this is so that the resulting module is scriptable. Otherwise, you can just have forward call the interpreter.

fx/native_interpreter/README.md
```diff
@@ -0,0 +1,29 @@
+# Lowering PyTorch Code to a Native Runtime With FX and TorchScript Custom Classes
+
+In this example, we are going to build an FX transformation takes the code from a PyTorch `Module` and converts (aka “lowers”) it to a format that a native C++ runtime can consume and execute. We are going to build up a trivial interpreter for this example, but you can imagine extending the same process to work with more sophisticated backends, ones which may do code optimization or offloading to an accelerator.
```
zdevito, line 3: I think there is more going on here than just a lowering, this example shows how to

1. Lower the FX graph to some other representation
2. Register that other representation as a custom operator that can be used from TorchScript or Python.
3. Wrap the lowered code into a format that can still be used within TorchScript.

Lots of things are simpler if you don't need to do 1--3, so a possible improvement would be to justify each component with whether it is for 1, 2, or 3 above.  I feel like without this clarification, this seems a lot more complicated than if someone, e.g., wanted to write a quick exporter to another format.

zdevito (approved):

jfix71 (approved):
Cool!

PR #896, opened 2021-04-02
fix the error of generated code in subgraph rewrite

XiaobingSuper:
The generated code is wrong, which shouldn't has `sum` op after running `replace_pattern`.


XiaobingSuper:
@jamesr66a

jamesr66a (approved):
cc @ansley

ansley:
Looks good! Thanks for catching this

PR #906, opened 2021-05-03
[RPC Framework] Create a hybrid model example using both DDP and RemoteModule

SciPioneer:
Replace `emb_rref` with `RemoteModule`. A key difference is that: the former was created on the master process and then passed to the trainers, and the latter was created on each of the trainers directly.


SciPioneer:
This changes the semantics of the example. After support passing `RemoteModule` as an arg over RPC (https://github.com/pytorch/pytorch/pull/57695), no semantics need to be changed.

PR #916, opened 2021-05-27
Fix comment to use zero-indexed ranks

andwgu:
The tutorial otherwise uses zero-indexed ranks throughout, so this change makes the example mentioned in the comment in `demo_basic()` follow the same convention.

No testing was performed since this only edits comments.


H-Huang (approved):
Thanks! changes are good

PR #945, opened 2021-12-29
Updating retracing examples to handle n-ary operators

zephirefaith:
Summary:
Replaced Proxy(object) creations with Proxy(object, GraphAppendingTracer)
which ensures graceful handling of n-ary operators.

By default Proxy creates a new tracer when no tracer arg
is provided, these examples encourage use of Proxy creation
with the GraphAppendingTracer when dealing with n-ary operators.

Implements cases outlined in pytorch/pytorch#68195

Test Plan:
```bash
python fx/inline_function.py
python fx/proxy_based_graph_creation.py
python fx/primitive_library.py
```

No errors on the above and output of `fx/primitive_library` confirmed
to be the same as inline commented expected output. Also verified  via
diff of OG output and updated output.

FYI: `fx/nnc_compile.py` might need this fix too, but I was not able to run
the script at all. Threw the following error:
`AttributeError: module 'torch._C._te' has no attribute 'KernelScope'`

Reviewers:
jamesreed
qianxuqx

Subscribers:

Tasks:
T105739531

Tags:

### DIFF
--- a/fx/inline_function.py
+++ b/fx/inline_function.py
@@ -40,10 +40,11 @@ class M(torch.nn.Module):
 # represented as a `call_module` Node. The full operation in the
 # generated `forward` function's code will appear as `self.relu(x)`
 m = symbolic_trace(M())
-tracer = torch.fx.proxy.GraphAppendingTracer(m.graph)
 
 # Insert nodes from the ReLU graph in place of the original call to
 # `self.relu`
+# create a graph-appending tracer pointing to the original graph
+tracer = torch.fx.proxy.GraphAppendingTracer(m.graph)
 for node in m.graph.nodes:
     # Find `call_module` Node in `m` that corresponds to `self.relu`.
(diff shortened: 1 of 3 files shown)


fx/inline_function.py
```diff
@@ -40,6 +40,7 @@ def forward(self, x):
 # represented as a `call_module` Node. The full operation in the
 # generated `forward` function's code will appear as `self.relu(x)`
 m = symbolic_trace(M())
+tracer = torch.fx.proxy.GraphAppendingTracer(m.graph)
```
jamesr66a, line 43: nit: IMO this should appear after the comment on line 46, since it is logically part of the modification process that starts there rather than the initial program capture up here

jamesr66a (approved):
Looks great, thank you!

fx/inline_function.py
zephirefaith: Resolved in b9071ca below.

PR #944, opened 2021-12-26
change torchtext to torchtext.legacy

yorkville101:
proposed fix for #942

According to Torchtext 0.9.0 release note:
- torchtext.data.Field -> torchtext.legacy.data.Field
- torchtext.data.Dataset -> torchtext.legacy.data.Dataset

https://github.com/pytorch/text/releases/tag/v0.9.0-rc5

Issue #943, opened 2021-11-24
CI Job Reports Success even when a Test Fails


mostafaelhoushi:
For example, in this CI [job](https://github.com/pytorch/examples/runs/4293561354?check_suite_focus=true) we see at the end of the log that one of the tests have failed:
```
Finished run_all, status 0
Starting clean
running clean to remove cruft
Updated 1 path from the index
Finished clean, status 0
Some examples failed:

couldn't train snli
```
but the CI has a green signal.

Probably, the solution is:
- First we need to fix that snli job (I created issue #942 for that).
- Then make `run_python_examples.sh` script return status -1 if a test fails. (See https://www.oreilly.com/library/view/solaristm-7-reference/0130200484/0130200484_ch05lev1sec13.html)

Issue #942, opened 2021-11-24
snli job is failing


mostafaelhoushi:
When running:
```
./run_python_examples.sh snli
```

we get:
```
Traceback (most recent call last):
  File "train.py", line 23, in <module>
    inputs = data.Field(lower=args.lower, tokenize='spacy')
AttributeError: module 'torchtext.data' has no attribute 'Field'
couldn't train snli
```

PR #947, opened 2022-01-14
Fix CI Job Report when a test fail #943

Kord96:
Fix [CI Job Reports Success even when a Test Fails #943
](https://github.com/pytorch/examples/issues/943). The suggested solution was

1. Fix [snli job is failing #942](https://github.com/pytorch/examples/issues/942) which was performed in [change torchtext to torchtext.legacy #944](https://github.com/pytorch/examples/pull/944)
2. make run_python_examples.sh exit with error status which is performed in this pull request.


mostafaelhoushi:
Thank you @Kord96 . The CI did report failure as it should after your change.
Now after I merged #944 , I am re-running CI to double check that it passes, then I will merge.

PR #949, opened 2022-02-08
#71844 Migrate default branch to main.

musebc:
This PR includes updates to change the default branch from master to main.

Fixes #71844 from pytorch repo.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,13 +2,13 @@ name: Run Examples
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
   schedule:
     # Every day at 3:00am
     - cron: '0 3 * * *'
-        
+
 
 jobs:
   test:
@@ -39,4 +39,3 @@ jobs:
         title: Daily CI failed
         body:  Commit ${{ github.sha }} daily scheduled [CI run](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}) failed, please check why
         assignees: ''
-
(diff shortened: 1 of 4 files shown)


janeyx99 (approved):
Looks good to me. Thanks for the change and sorry for not seeing it earlier!

janeyx99:
Hey @nairbv, this looks good to me, but would appreciate a glance from you before @musebc merges this/changes the default branch internally.

.github/workflows/main.yml
```diff
@@ -2,13 +2,13 @@ name: Run Examples
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
janeyx99, line 5: Maybe we should do that thing where we include both master and main here for the smoothest transition?

nairbv (commented):
I'm fine with the changes but should we actually make the change on `main` instead of `master`?

i.e. we could leave master as-is, make these change on `main`, and update the default branch to be `main`.

musebc:
@nairbv We could do that, but the flow we have done before is to rename the `master` branch to `main` using the internal tools. This also migrates the branch protections which is helpful.

.github/workflows/main.yml
musebc: Actually, I just had a thought about this. When we migrate the branches, the only one that will be protected is main and then we could have the reverse issue as in the pytorch/pytorch. We either need to protect both, or just leave this be and merge/change the branch name together

musebc: I think it's not a big deal on this repo so I could go either way, but I think it might just be easiest to just keep it as is and do the swap quickly.

nairbv (approved):

Issue #933, opened 2021-09-19
Imagenet example shell script is only for validation


MaxVanDijck:
https://github.com/pytorch/examples/tree/master/imagenet

The shell script specified in the readme only sorts the images of the validation set into folders
There is no script to extract the training set included in the readme
I suggest including this https://gist.github.com/BIGBALLON/8a71d225eff18d88e469e6ea9b39cef4 in the readme to extract and prep both the training and validation .tar files

msaroufim:
Hi @MaxVanDijck would you like to make a PR for this?

MaxVanDijck:
Yes, I can update the link in the readme today

MaxVanDijck:
PR addressing issue: https://github.com/pytorch/examples/pull/977

PR #977, opened 2022-03-11
Change imagenet data script to include training data

MaxVanDijck:
Added script in readme to manage extraction of both training and validation data in reference to: https://github.com/pytorch/examples/issues/933

### DIFF
--- a/imagenet/README.md
+++ b/imagenet/README.md
@@ -7,7 +7,7 @@ This implements training of popular model architectures, such as ResNet, AlexNet
 - Install PyTorch ([pytorch.org](http://pytorch.org))
 - `pip install -r requirements.txt`
 - Download the ImageNet dataset from http://www.image-net.org/
-    - Then, and move validation images to labeled subfolders, using [the following shell script](https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh)
+    - Then, move and extract the training and validation images to labeled subfolders, using [the following shell script](https://gist.github.com/MaxVanDijck/39531343d1c2ff6a0ee3304d8dee917f)
 
 ## Training
 
--- a/imagenet/README.md
+++ b/imagenet/README.md
@@ -7,7 +7,7 @@ This implements training of popular model architectures, such as ResNet, AlexNet
 - Install PyTorch ([pytorch.org](http://pytorch.org))
 - `pip install -r requirements.txt`
 - Download the ImageNet dataset from http://www.image-net.org/
-    - Then, move and extract the training and validation images to labeled subfolders, using [the following shell script](https://gist.github.com/MaxVanDijck/39531343d1c2ff6a0ee3304d8dee917f)
+    - Then, move and extract the training and validation images to labeled subfolders, using [the following shell script](extract_ILSVRC.sh)
 
 ## Training
 
--- a/imagenet/extract_ILSVRC.sh
+++ b/imagenet/extract_ILSVRC.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# script to extract ImageNet dataset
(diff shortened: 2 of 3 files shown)


imagenet/README.md
```diff
@@ -7,7 +7,7 @@ This implements training of popular model architectures, such as ResNet, AlexNet
 - Install PyTorch ([pytorch.org](http://pytorch.org))
 - `pip install -r requirements.txt`
 - Download the ImageNet dataset from http://www.image-net.org/
-    - Then, and move validation images to labeled subfolders, using [the following shell script](https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh)
+    - Then, move and extract the training and validation images to labeled subfolders, using [the following shell script](https://gist.github.com/MaxVanDijck/39531343d1c2ff6a0ee3304d8dee917f)
```
msaroufim, line 10: Would suggest adding your script to source control since it's pretty readable and could you add some brief explanation or proof as to how your script doesnt change the intended behavior in soumith's old script?

msaroufim (changes requested):

imagenet/README.md
MaxVanDijck: previously you have to extract the validation .tar file inside a directory named 'val' and then run Soumith's script

Soumith script:
(creates directories for categories)
`mkdir -p n01440764`
(moves images into directories)
`mv ILSVRC2012_val_00000293.JPEG n01440764/`

this results in 
val/
├── n01440764
│   ├── ILSVRC2012_val_00000293.JPEG
│   ├── ......
├── ......



New script uses extraction method from archive here modified to extract images to 'imagenet/train' and 'imagenet/val'
https://github.com/facebookarchive/fb.resnet.torch/blob/master/INSTALL.md#download-the-imagenet-dataset

for training data:
create directory, move .tar file and change directory to imagenet/train
`mkdir imagenet/train && mv ILSVRC2012_img_train.tar imagenet/train/ && cd imagenet/train`
extract training set into dir and remove compressed file
`tar -xvf ILSVRC2012_img_train.tar && rm -f ILSVRC2012_img_train.tar`

this results in a directory of compressed .tar files, one for each folder of images per each category

find all .tar files, create directory with same name of tar file, decompress and copy contents of tar file to new directory, remove .tar file
`find . -name "*.tar" | while read NAME ; do mkdir -p "${NAME%.tar}"; tar -xvf "${NAME}" -C "${NAME%.tar}"; rm -f "${NAME}"; done`

this results in a directory full of new directories, each with the same name of the .tar file they were created from and containing the contents of the .tar file in this case images
eg.
train/
├── n01440764
│   ├── n01440764_10026.JPEG
│   ├── n01440764_10027.JPEG
│   ├── ......
├── ......

for the validation data:
create directory, move .tar file, change directory to imagenet/val and extract validation data
`mkdir imagenet/val && mv ILSVRC2012_img_val.tar imagenet/val/ && cd imagenet/val && tar -xvf ILSVRC2012_img_val.tar`
then simply wget and run Soumith's script for validation data
`wget -qO- https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh | bash`

resulting in 
val/
├── n01440764
│   ├── ILSVRC2012_val_00000293.JPEG
│   ├── ......
├── ......


**TLDR** In essence, validation set is managed in the same way except you don't have to extract before running Soumith's script
train set is managed using the names of the .tar files for each category after the initial ILSVRC2012_img_train.tar is extracted

msaroufim: Nice do you mind summarizing this thread in the README for this example or in your shell script?

msaroufim (changes requested):
Let's just integrate the notes from your comment into either the README or the shell script to explain this change

msaroufim (approved):

Issue #887, opened 2021-02-14
OSError: [Errno 22] Invalid argument:


Feez:
https://github.com/pytorch/examples/blob/36441a83b6595524a538e342594ee6482754f374/fast_neural_style/neural_style/neural_style.py#L117

Spent 2 hours training a model for it to fail the save with error prefixed with `OSError: [Errno 22] Invalid argument` :(

For now I'm just adding fail-safe but I am sure it can be written cleaner
```py
custom_save_path = save_model_path
    while True:
        try:
            torch.save(transformer.state_dict(), custom_save_path)
            break
        except:
            print("Please enter new save path:")
            custom_save_path = input() 

    print("\nDone, trained model saved at", custom_save_path)
```

msaroufim:
@Feez would you like to make a PR for this? Could you also show us the full error you're getting?

Feez:
I'm sorry I don't have this environment set up anymore but I believe the error was a result of just inserting a save path that is syntactically valid but didn't actually exist. Such as a save path in a directory that doesn't exist.

Regardless I think there probably should be more fail safes around the actual saving of the file to prevent loss of progress in training a model, as I imagine this could happen with many other OS related issues that the user might want to manually resolve and attempt to re-save (e.g. out of disk space on save path, drive unplugged during save, network path disconnected, etc)

Feez:
Created a quick PR to address the general problem stated above. Hopefully this suffices.

PR #979, opened 2022-03-12
Fail safe on fast_neural_style transformer file saving

Feez:
Related: https://github.com/pytorch/examples/issues/887


msaroufim:
This seems like a generally useful pattern, maybe worth having this be argument for a fault tolerant `torch.save()` - I'm make the request and elaborate the need on `pytorch/pytorch`

PR #981, opened 2022-03-16
update vmap.py to use current torch.fx interface

constroy:


msaroufim:
@constroy @Chillee could we add a test for fx stuff here https://github.com/pytorch/examples/blob/main/run_python_examples.sh

msaroufim:
Hi @constroy I apologize for the delay but I ended up deprecating this specific example because I figured it may be better to point folks to `pytorch/functorch`

Issue #976, opened 2022-03-11
Update all examples to latest versions of pytorch


msaroufim:
Our RPC examples use an older version of PyTorch 

```
grep -r "torch=" .
./distributed/rpc/pipeline/requirements.txt:torch==1.6.0
./distributed/rpc/batch/requirements.txt:torch==1.6.0
```

There could be others

veekaybee:
I'd like to take this. Just to clarify, we should update all requirements files (starting [with here](https://github.com/pytorch/examples/blob/main/distributed/rpc/pipeline/requirements.txt), to `torch==1.9.0`. 

I did a quick check of 
```
find . -name '*requirements.txt' -exec grep -i 'torch' {} \; -print
```

and it looks like there are others that use torch but don't pin a version. Does it make sense to pin all of them and then do some kind of automated CI/CD check? Maybe that's phase 2 or 3. :)

Sagnik2003:
@msaroufim I would like to work on it.

PR #988, opened 2022-04-12
Requirements upgrades and clarification in CONTRIBUTING

veekaybee:
Addressing this issue: https://github.com/pytorch/examples/issues/976 and changing all the versions of PyTorch where dependencies were pinned to 1.9.0 and also clarifying install instructions for contributing. 

```
find . -name '*requirements.txt' -exec grep -i 'torch==' {} \; -print
torch==1.9.0
./distributed/rpc/pipeline/requirements.txt
torch==1.9.0
./distributed/rpc/ddp_rpc/requirements.txt
torch==1.9.0
./distributed/rpc/batch/requirements.txt
```

All examples in the build script build and run correctly but I did have to make some tweaks to `DataLoader` to get them to work on my Mac.


distributed/rpc/ddp_rpc/requirements.txt
```diff
@@ -1 +1 @@
-torch>=1.6.0
+torch==1.9.0
```
msaroufim, line 1: Can you double-check if we can reenable this test https://github.com/pytorch/examples/blob/main/run_python_examples.sh#L176 and add a small function for it

May have a couple of APIs that have been deprecated so worst case we can just make a documentation only PR

CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ If you're new we encourage you to take a look at issues tagged with [good first
 ## For bug fixes
 1. Fork the repo and create your branch from `main`.
 2. Make your fix
-3. Make sure that `run_python_examples.sh` passes locally
+3. Make sure that `run_python_examples.sh` passes locally by first installing dependencies with `./run_python_examples.sh "install_deps"` then running the script end to end.
```
msaroufim, line 22: Maybe we can create a separate step for installing dependencies

msaroufim (changes requested):

distributed/rpc/ddp_rpc/requirements.txt
veekaybee: I was able to create tests to trigger the method, but because my local machine is a Macbook I don't have access to CUDA. What would you recommend as the best way to test it - my hunch is that it would make most sense to spin up a deep learning image in any given cloud. Thoughts?

msaroufim: That's what I do, I usually use one the Ubuntu Deep Learning AMI on AWS, you can choose either a p2 or g4dn instance

msaroufim: @veekaybee I'm super sorry but now that I think more about this we'd need to also have a GPU instance in a Github runner as well to run GPU tests on CI so there's no need to add tests but if you could sanity check that that the scripts do run indeed once you upgrade the torch version, maybe all the way to 1.11 then this would still be a valuable contribution

veekaybee: I'm getting the following error when testing running the [main method here](https://github.com/pytorch/examples/blob/349c08e95acda442978e83400648afb3e3764036/distributed/rpc/ddp_rpc/main.py#L177) 

```
Running example: distributed
On WorkerInfo(id=1, name=trainer1):
RuntimeError('CUDA error: invalid device ordinal\nCUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.\nFor debugging consider passing CUDA_LAUNCH_BLOCKING=1.')
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/distributed/rpc/internal.py", line 204, in _run_function
    result = python_udf.func(*python_udf.args, **python_udf.kwargs)
  File "/home/ubuntu/examples/distributed/rpc/ddp_rpc/main.py", line 48, in _run_trainer
    model = HybridModel(remote_emb_module, rank)
  File "/home/ubuntu/examples/distributed/rpc/ddp_rpc/main.py", line 30, in __init__
    self.fc = DDP(torch.nn.Linear(16, 8).cuda(device), device_ids=[device])
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 688, in cuda
    return self._apply(lambda t: t.cuda(device))
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 601, in _apply
    param_applied = fn(param)
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 688, in <lambda>
    return self._apply(lambda t: t.cuda(device))
RuntimeError: CUDA error: invalid device ordinal
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

[W tensorpipe_agent.cpp:682] RPC agent for ps encountered error when reading incoming request from master: eof (this error originated at tensorpipe/transport/shm/connection_impl.cc:259)
[W tensorpipe_agent.cpp:682] RPC agent for trainer1 encountered error when reading incoming request from master: eof (this error originated at tensorpipe/transport/shm/connection_impl.cc:259)
[W tensorpipe_agent.cpp:682] RPC agent for trainer0 encountered error when reading incoming request from master: eof (this error originated at tensorpipe/transport/shm/connection_impl.cc:259)
[W tensorpipe_agent.cpp:682] RPC agent for ps encountered error when reading incoming request from trainer0: eof (this error originated at tensorpipe/transport/shm/connection_impl.cc:259)
[W tensorpipe_agent.cpp:682] RPC agent for ps encountered error when reading incoming request from trainer1: eof (this error originated at tensorpipe/transport/shm/connection_impl.cc:259)
  File "main.py", line 180, in <module>
    mp.spawn(run_worker, args=(world_size,), nprocs=world_size, join=True)
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 240, in spawn
    return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 198, in start_processes
    while not context.join():
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 160, in join
    raise ProcessRaisedException(msg, error_index, failed_process.pid)
torch.multiprocessing.spawn.ProcessRaisedException: 

-- Process 2 terminated with the following error:
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/site-packages/torch/multiprocessing/spawn.py", line 69, in _wrap
    fn(i, *args)
  File "/home/ubuntu/examples/distributed/rpc/ddp_rpc/main.py", line 146, in run_worker
    fut.wait()
RuntimeError: RPCErr:1:RPC ran for more than set timeout (60000 ms) and will now be marked with an error
```

Based on looking at the code, it's because the [world size needs to be 4.](https://pytorch.org/docs/stable/rpc.html) Does this assume four separate machines that each map one process, or 4 distinct processes in one machine?  

Should I change the logic in [run_worker](https://github.com/pytorch/examples/blob/349c08e95acda442978e83400648afb3e3764036/distributed/rpc/ddp_rpc/main.py#L108) to assume one machine or is the intended behavior to work out of the box?

msaroufim: world size of 4 means 4 separate processes which could either be on the same or different machines. They typically get gang scheduled so if one fails they all fail. So ideally `run_worker` shouldn't assume a local only scenario but if that's simplest for you to test I'm still happy merging the PR because it's not like we have tests for either case running now

msaroufim (changes requested):

distributed/rpc/ddp_rpc/requirements.txt
veekaybee: Just a quick update on this: I'm running these tests on 1.9 but encountering some issues that are taking a while to resolve, likely due to [these API updates]( https://discuss.pytorch.org/t/rpc-behavior-difference-between-pytorch-1-7-0-vs-1-9-0/124772/7). 

So far, I'm trying by tuning `run_worker` options like so, by tuning ulimits on my EC2 instance, and through increasing the timeout. 

`rpc_backend_options = TensorPipeRpcBackendOptions(_transports=["uv"])`

veekaybee: It doesn't look like we can re-enable this test for 1.9 quite yet.

msaroufim: That's ok maybe we merge your documentation change and we can have a seperate PR to enable the distributed test

veekaybee: Sure, just for these correct `distributed/rpc/ddp_rpc/requirements.txt`? The other Makefile steps pass.

msaroufim: yup that's right

veekaybee: Ok, I reverted that specific change, let me know if it's ok to go otherwise. I can also file a new issue specific to this and assign it to myself. I'll use the previous issues as examples of how to write one.

veekaybee: Created a new issue to follow up on this here https://github.com/pytorch/examples/issues/991

Issue #862, opened 2020-12-29
Why not move images onto gpu?


I-Doctor:
https://github.com/pytorch/examples/blob/792d336019a28a679e29cf174e10cee80ead8722/imagenet/main.py#L284

I'm trying to training vgg on imagenet with one node DataParallel and no multiprocessing。But I find 'images.device' before computation is 'cpu', and I'm not sure why these codes move 'images' to gpu only when I choose only one gpu and move 'targegt' to gpu even with argument device=None(args.gpu=None).

suraj813:
Thanks @I-Doctor, [this is](https://github.com/pytorch/examples/blob/792d336019a28a679e29cf174e10cee80ead8722/imagenet/main.py#L284-L287) a weird snippet that has multiple occurrences in this script. The second condition (if `torch.cuda.is_available()`) can be merged into the first (`if args.gpu is not None`).  Would you like to raise a PR for this?

soumith:
what @suraj813 suggested is incorrect.

>  I'm not sure why these four lines of codes move 'images' to gpu only when I choose only one gpu(args.gpu is not None) and move 'target' to gpu even with argument device=None(args.gpu=None).


the reason is this.
When you do single GPU training, you dont use `nn.DataParallel`, you simply send `images` directly to your model. So `images` is moved onto GPU 0 in a straight-forward manner.

When you do multi-GPU training with `nn.DataParallel`, you keep `images` on CPU and then send it to `nn.DataParallel`, which will slice `images` into as many Tensors as there are GPUs (for example if `images` had 64 images and there  are 4 GPUs, `images` is sliced into four separate Tensors of `16` images each). These individual sliced Tensors are then sent to their respective GPUs. In the 4-GPU example, if we had sent `images` onto GPU 0 and then called `nn.DataParallel` there would be a wasteful move of `images` first to GPU0, and then each of the 3 slices of images will then each be moved from GPU0 to GPU1, GPU0 to GPU2 and GPU0 to GPU3 respectively.

suraj813:
Thanks @soumith for clarifying what that `if` block is doing. IIUC the two conditions are to cater to both `nn.DataParallel` and `DistributedDataParallel`

When `args.gpu == None` i.e. we are doing multi-GPU training with `nn.DataParallel`,
 * `images` are kept on CPU and not moved to any GPU 
 * `targets` is moved to GPU0 (but this will instead be the specific GPU when we are using `DistributedDataParallel`)
 
[...]

PR #995, opened 2022-04-23
Fixed distributed GPU bug in ImageNet example

Quentin-Anthony:
The current imagenet example doesn't move images to the GPU in the distributed GPU case, only in the single GPU case:

https://github.com/pytorch/examples/blob/main/imagenet/main.py#L293
https://github.com/pytorch/examples/blob/main/imagenet/main.py#L337

This leads to errors like the following:

```
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```

I think this can be fixed by relaxing the `if` statement.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -290,9 +290,9 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         # measure data loading time
         data_time.update(time.time() - end)
 
-        if torch.cuda.is_available():
+        if args.gpu is not None:
             images = images.cuda(args.gpu, non_blocking=True)
-        if torch.cuda.is_available():
+        if args.gpu is not None:
             target = target.cuda(args.gpu, non_blocking=True)
 
         # compute output
@@ -334,9 +334,9 @@ def validate(val_loader, model, criterion, args):
     with torch.no_grad():
         end = time.time()
         for i, (images, target) in enumerate(val_loader):
-            if torch.cuda.is_available():
+            if args.gpu is not None:
                 images = images.cuda(args.gpu, non_blocking=True)
-            if torch.cuda.is_available():
+            if args.gpu is not None:
                 target = target.cuda(args.gpu, non_blocking=True)
 
             # compute output
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -290,7 +290,7 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         # measure data loading time
         data_time.update(time.time() - end)
 
-        if args.gpu is not None:
+        if torch.cuda.is_available():
             images = images.cuda(args.gpu, non_blocking=True)
         if torch.cuda.is_available():
             target = target.cuda(args.gpu, non_blocking=True)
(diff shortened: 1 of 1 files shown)


imagenet/main.py
```diff
@@ -290,7 +290,7 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
         # measure data loading time
         data_time.update(time.time() - end)
 
-        if args.gpu is not None:
+        if torch.cuda.is_available():
```
soumith, line 293: i think this should be args.gpu should be existing with regular distributed.
Do you mean it's not working in multiprocessing distributed mode?
https://github.com/pytorch/examples/blob/main/imagenet/main.py#L73-L77

Quentin-Anthony: The opposite actually. Running with ```--multiprocess-distributed``` sets ```args.gpu``` here: https://github.com/pytorch/examples/blob/main/imagenet/main.py#L119

Running regular distributed (in my case by setting the ```--world-size```, which triggers ```args.distributed``` here: https://github.com/pytorch/examples/blob/main/imagenet/main.py#L102)  leaves ```args.gpu``` unset. 

It looks like this issue has existed for some time, since I found this issue on it with the same suggested fix: https://github.com/pytorch/examples/issues/862

soumith: If I remember correctly (and I looked at the code more closely now), when you launch distributed yourself (i.e. without relying on multiprocessing distributed), you are responsible for giving `--world-size` and `--gpu` correctly.

For example, the example usage would be:

```
python main.py ... --world-size 2 --rank 0 --gpu 0 imagenet_root &
python main.py ... --world-size 2 --rank 1 --gpu 1 imagenet_root &
```

The patch as you've written it will always effectively run `images = images.cuda(None, non_blocking=True)`, putting the images on GPU0 regardless of which GPU it is supposed to be running on.

soumith: I think the suggested fix in https://github.com/pytorch/examples/issues/862 is wrong. I'll comment there as well.

Quentin-Anthony: I see. Thanks for the explanation! Can I create a PR with this example usage in the README so that other users launching regular distributed don't face this issue? 

**Last question:** Would it be helpful for readability to normalize these ```if``` statements? If so, I've updated the PR for this. If not, we can reject this PR entirely.

imagenet/main.py
```diff
@@ -292,7 +292,7 @@ def train(train_loader, model, criterion, optimizer, epoch, args):
 
         if args.gpu is not None:
             images = images.cuda(args.gpu, non_blocking=True)
-        if torch.cuda.is_available():
+        if args.gpu is not None:
```
suraj813, line 295: Changing this condition will break `nn.DataParallel` because the DataParallel model's output will be on GPU0. This situation is (hopefully) clarified in https://github.com/pytorch/examples/issues/862#issuecomment-1111118888

imagenet/main.py
soumith: > Can I create a PR with this example usage in the README so that other users launching regular distributed don't face this issue?

Yes

PR #1001, opened 2022-04-30
Add distributed usage to ImageNet example's README

Quentin-Anthony:
Following the discussions in #995 and #862, This PR adds a small usage example into the ImageNet README.

One note, however, is that specifying ```--gpu``` will trigger the print in https://github.com/pytorch/examples/blob/main/imagenet/main.py#L95-L97, which is confusing to users. Should we either remove or clarify this print?

### DIFF
--- a/imagenet/README.md
+++ b/imagenet/README.md
@@ -45,6 +45,15 @@ Node 1:
 python main.py -a resnet50 --dist-url 'tcp://IP_OF_NODE0:FREEPORT' --dist-backend 'nccl' --multiprocessing-distributed --world-size 2 --rank 1 [imagenet-folder with train and val folders]
 ```
 
+## Distributed Data Parallel Training
+
+If you wish to disable PyTorch's multiprocessing module and manually manage processes yourself (e.g. with MPI), you must specify the `world-size`, `rank` and `gpu` values yourself. For example:
+
+```bash
+python main.py ... --world-size 2 --rank 0 --gpu 0 [imagenet-folder with train and val folders] &
+python main.py ... --world-size 2 --rank 1 --gpu 1 [imagenet-folder with train and val folders] &
+```
+
 ## Usage
 
 ```


imagenet/README.md
```diff
@@ -45,6 +45,15 @@ Node 1:
 python main.py -a resnet50 --dist-url 'tcp://IP_OF_NODE0:FREEPORT' --dist-backend 'nccl' --multiprocessing-distributed --world-size 2 --rank 1 [imagenet-folder with train and val folders]
 ```
 
+## Distributed Data Parallel Training
+
+If you wish to disable PyTorch's multiprocessing module and manually manage processes yourself (e.g. with MPI), you must specify the `world-size`, `rank` and `gpu` values yourself. For example:
+
+```bash
+python main.py ... --world-size 2 --rank 0 --gpu 0 [imagenet-folder with train and val folders] &
+python main.py ... --world-size 2 --rank 1 --gpu 1 [imagenet-folder with train and val folders] &
```
hudeven, line 54: @Quentin-Anthony noob question: why do you need the trailing "&"?

hudeven:
I think the print for `--gpu` is useful. To reduce the confusion, we can add some clarification to the print message. Alternative: as DDP is recommended to replace DataParallel, we should probably remove the usage of DataParallel in examples and update the print message accordingly.

imagenet/README.md
Quentin-Anthony: @hudeven -- It launches the line in the background so that bash can run the next command immediately

Quentin-Anthony:
> I think the print for `--gpu` is useful. To reduce the confusion, we can add some clarification to the print message. Alternative: as DDP is recommended to replace DataParallel, we should probably remove the usage of DataParallel in examples and update the print message accordingly.

I'm confused. This PR is for updating the README. What's being printed?

PR #1011, opened 2022-05-27
add dummy data benchmark

KimBioInfoStudio:


KimBioInfoStudio:
@msaroufim could u help assign some people help on code review?

msaroufim (changes requested):
Thank you for your contributions just trying to understand the full scope there seems to be some
* linting changes in which case could you specify how you linted?
* `[INFO]` before some print statements, not sure if this is consistent with the rest of the repo
* Because imagenet is hard to download you've added support for a `FakeData` class
* Some new documentation where you use `--data` instead of `--dummy`?
* You expanded support for new architectures dramatically, did you test them all?

I'd appreciate reading a PR description to better understand what this change is about and document in the long term the benefits to users. I'd also recommend just making the change to support `--dummy` data first and then we can do another PR where we integrate a Github Action based linter

KimBioInfoStudio:
> Q: * linting changes in which case could you specify how you linted?
A: use  autopep8 to lint the python code and max-line-length is 120
> * `[INFO]` before some print statements, not sure if this is consistent with the rest of the repo
A: if this is not acceptable this time I could remove the change
> * Because imagenet is hard to download you've added support for a `FakeData` class which is a great change
> * Some new documentation where you use `--data` instead of `--dummy`?
A: original  `data` is a position arg, we could NOT set it when using `--dummy`, so change `data` to an optional arg
> * You expanded support for new architectures dramatically, did you test them all?
A: yep, `--dummy` was performed with x86 CPU
> 
> I'd appreciate reading a PR description to better understand what this change is about and document in the long term the benefits to users. I'd also recommend just making the change to support `--dummy` data first and then we can do another PR where we integrate a Github Action-based linter

PR #1014, opened 2022-06-15
[PT-D] Add TP and DDP into the example run script

fduwjj:


msaroufim:
Hi @fduwjj this repo doesn't have GPU runners right now, is it possible to have a test run for a single batch on CPU just as a sanity check? https://github.com/pytorch/examples/runs/6908535986?check_suite_focus=true#step:5:627

Otherwise I think we can consider adding GPU runners for examples but this would be something new - maybe @svekars or @seemethere have some thoughts here

fduwjj:
Hi, @msaroufim. Currently, tensor parallel is only supported on CUDA because we need to use NCCL. So we cannot have a CPU version. I think down the road, PyTorch Distributed and other PyTorch components will have more GPU based library or features enabled. So it's worth considering to have a GPU runner.

msaroufim (approved):

PR #1017, opened 2022-06-30
env.seed is deprecated, change to env.reset

rogerioagjr:
env.seed is deprecated and this PR updates that to env.reset, which gets rid of the warning message.


msaroufim (approved):
LGTM, thank you!

Issue #1023, opened 2022-07-14
parameter server example not running in ubuntu


gagandeep987123:
I am running the server and worker in separate sessions but execution stalls in the worker thread at line 175 while running 
`    for i, (data, target) in enumerate(train_loader):
`
However, it is running fine on Mac. 
The issue only happens in ubuntu.
Can you please help

hudeven:
@gagandeep987123 thanks for reporting the issue! It seems to be deadlock caused by "fork" start method(default in Linux). In Mac, the default is "spawn" so it works fine. You could try to add "mp.set_start_method("spawn")" before any process starts.

hudeven:
@gagandeep987123 Please feel free to reopen this issue if it's not resolved by https://github.com/pytorch/examples/pull/1025

PR #1025, opened 2022-07-18
[parameter_server] fix deadlock in multiprocessing by using spawn

hudeven:
Bug fix for https://github.com/pytorch/examples/issues/1023

Tested it in Mac with CPU, in Linux with CPU and 2 GPUs(aws p3.8xlarge instance with Amazon Linux/RHEL).


tangbinh (approved):
LGTM!

PR #1028, opened 2022-08-02
refine imagenet dummy part of #1011

KimBioInfoStudio:
we add a method to use dummy data for performance benchmark only, this feature will save a lot of time for downloading real ImageNet Dataset, this is a part of #1011 

image

### DIFF
--- a/imagenet/README.md
+++ b/imagenet/README.md
@@ -23,6 +23,14 @@ The default learning rate schedule starts at 0.1 and decays by a factor of 10 ev
 python main.py -a alexnet --lr 0.01 [imagenet-folder with train and val folders]
 ```
 
+## Dummy  Data for Performance Benchmark Only
+
+Due to the data is dummy only for Performance Benchmark, the loss/accuracy is not reliable.
+
+```bash
+python main.py -a resnet18 --dummy
+```
+
 ## Multi-processing Distributed Data Parallel Training
 
 You should always use the NCCL backend for multi-processing distributed training since it currently provides the best distributed training performance.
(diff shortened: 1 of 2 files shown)


KimBioInfoStudio:
@msaroufim could u help review again

msaroufim:
Tagging @hudeven

imagenet/README.md
```diff
@@ -23,6 +23,14 @@ The default learning rate schedule starts at 0.1 and decays by a factor of 10 ev
 python main.py -a alexnet --lr 0.01 [imagenet-folder with train and val folders]
 ```
 
+## Dummy  Data for Performance Benchmark Only
+
+Due to the data is dummy only for Performance Benchmark, the loss/accuracy is not reliable.
```
hudeven, line 28: I think it's not only useful for training speed benchmark but also for quick get started. How about rephrasing it like below?

```suggestion
## Use Dummy Data

ImageNet dataset is large and time-consuming to download. To get started quickly, run `main.py` using dummy data by "--dummy". It's also useful for training speed benchmark. Note that the loss or accuracy is useless in this case.
```

imagenet/main.py
```diff
@@ -206,28 +207,33 @@ def main_worker(gpu, ngpus_per_node, args):
     cudnn.benchmark = True
 
     # Data loading code
-    traindir = os.path.join(args.data, 'train')
-    valdir = os.path.join(args.data, 'val')
-    normalize = transforms.Normalize(mean=[0.485, 0.456, 0.406],
+    if args.dummy:
+        print("=> Dummy data is using!")
```
hudeven, line 211: nit 
```suggestion
        print("=> Dummy data is used!")
```

hudeven (commented):
@KimBioInfoStudio thanks for your contribution! Overall LGTM. I left some comments about documentation.

hudeven (approved):
LGTM.

PR #1031, opened 2022-08-05
total must be CUDA

JoohyungLee0106:
Currently, ``` total ```(https://github.com/pytorch/examples/blob/main/imagenet/main.py#L419) is at cpu but it must be CUDA to compute ``` dist.all_reduce() ```. Otherwise it will raise a ``` RuntimeError: Tensors must be CUDA and dense ```

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -415,7 +415,7 @@ class AverageMeter(object):
         self.avg = self.sum / self.count
 
     def all_reduce(self):
-        total = torch.FloatTensor([self.sum, self.count])
+        total = torch.FloatTensor([self.sum, self.count]).cuda(self.sum.device)
         dist.all_reduce(total, dist.ReduceOp.SUM, async_op=False)
         self.sum, self.count = total.tolist()
         self.avg = self.sum / self.count


imagenet/main.py
```diff
@@ -415,7 +415,7 @@ def update(self, val, n=1):
         self.avg = self.sum / self.count
 
     def all_reduce(self):
-        total = torch.FloatTensor([self.sum, self.count])
+        total = torch.FloatTensor([self.sum, self.count]).cuda(self.sum.device)
```
soumith, line 418: i think it should be `total = torch.Tensor([self.sum, self.count], device=self.sum.device, dtype=torch.float32)`

JoohyungLee0106: > i think it should be `total = torch.Tensor([self.sum, self.count], device=self.sum.device, dtype=torch.float32)`


That should work too :)

hudeven: Both works. I think " torch.tensor(...)" is more recommended nowadays. @JoohyungLee0106 would you like to make the change?

JoohyungLee0106: @hudeven Sure.

xwuShirley:
It seems  the current update https://github.com/pytorch/examples/blob/main/imagenet/main.py#L424 :(
falls back to the wrong version.....

I was trying to fix this bug and fortunately found this. It would be great it's updated 
total = torch.FloatTensor([self.sum, self.count]).cuda(self.sum.device)

hudeven:
As a similar change is already merged in https://github.com/pytorch/examples/pull/1034 , I'm closing this one.

PR #1034, opened 2022-08-16
Fix for NCCL backend, all_reduce performed on GPU when available

jataylo:
Use of torch.FloatTensor causes an issue with the NCCL backend, which only supports DistributedDataParallel primitives on GPU. This results in a runtime error "Tensors must be CUDA and dense".

This commit implemented a condition which checks if torch.cuda.is_available() and if so instead uses torch.cuda.FloatTensor to resolve the runtime error for the NCCL backend.

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -421,7 +421,10 @@ class AverageMeter(object):
         self.avg = self.sum / self.count
 
     def all_reduce(self):
-        total = torch.FloatTensor([self.sum, self.count])
+        if (torch.cuda.is_available()):
+            total = torch.cuda.FloatTensor([self.sum, self.count])
+        else:
+            total = torch.FloatTensor([self.sum, self.count])
         dist.all_reduce(total, dist.ReduceOp.SUM, async_op=False)
         self.sum, self.count = total.tolist()
         self.avg = self.sum / self.count
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -421,10 +421,8 @@ class AverageMeter(object):
         self.avg = self.sum / self.count
 
     def all_reduce(self):
-        if (torch.cuda.is_available()):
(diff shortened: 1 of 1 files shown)


imagenet/main.py
```diff
@@ -421,7 +421,10 @@ def update(self, val, n=1):
         self.avg = self.sum / self.count
 
     def all_reduce(self):
-        total = torch.FloatTensor([self.sum, self.count])
+        if (torch.cuda.is_available()):
+            total = torch.cuda.FloatTensor([self.sum, self.count])
+        else:
+            total = torch.FloatTensor([self.sum, self.count])
```
hudeven, line 427: Your implementation is correct. But how about using the modern syntax below?

```suggestion
        device = "cuda" if torch.cuda.is_available() else "cpu"
        total = torch.tensor([self.sum, self.count], dtype=torch.float32, device=device)
```

hudeven (changes requested):
Hi @jataylo , thanks for your contribution! As we don't have CI test for distributed training yet, would you mind pasting your CLI and output before/after this change in the description?

imagenet/main.py
jataylo: Good idea to make use of the modern syntax, its a lot cleaner than using the two different constructors. Have commited the suggestion.

jataylo:
CLI/Output supplied

**Output log files**
[original-logs.txt](https://github.com/pytorch/examples/files/9355445/original-logs.txt)
[fixed-logs.txt](https://github.com/pytorch/examples/files/9355446/fixed-logs.txt)

**CLI screenshots**

_before changes:_
bad-logs1
bad-logs2

_after changes:_
good-log-1
good-log-2
good-log-3

hudeven (approved):

Issue #1040, opened 2022-08-23
In example DCGAN, curl timed out 


ShiboXing:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/pytorch/serve/examples) before creating one.

## Context

* Pytorch version: 1.12.1
* Operating System and version: 20.04.4 LTS (Focal Fossa)

## Your Environment

* Installed using source? [yes/no]: no
* Are you planning to deploy it using docker container? [yes/no]: yes
* Is it a CPU or GPU environment?: GPU
* Which example are you using: DCGAN
* Link to code or data to repro [if any]:

## Expected Behavior

dcgan finishes without errors

## Current Behavior

dcgan fails with 

## Steps to Reproduce

1. cd examples
2. bash run_python_examples.sh "install_deps, dcgan"
...

## Failure Logs [if any]

```
Downloading classroom train set
--
181 | curl: /opt/conda/lib/libcurl.so.4: no version information available (required by curl)
182 | % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
183 | Dload  Upload   Total   Spent    Left  Speed
...
curl: (18) transfer closed with 3277022655 bytes remaining to read

...
Some examples failed:

couldn't unzip classroom
```

I know this is a thrid-party repo issue, which I have already raised in [lsun repo](https://github.com/fyu/lsun/issues/46)
Is it possible that you could have a solution on your end? The request speed of the domain http://dl.yf.io is just very slow in general.

Thank you!

hudeven:
@ShiboXing thanks for reporting the issue! Let's see if we can get any solution from https://github.com/fyu/lsun/issues/46 

Before that, I'm fixing the flaky test by using fake data in #1044

PR #1044, opened 2022-08-23
Fix flaky dcgan test by using fake data

hudeven:
The dcgan test is flaky as reported in https://github.com/pytorch/examples/issues/1038 and https://github.com/pytorch/examples/issues/1040 
The root cause is that it's very slow to download the "classroom" dataset from http://dl.yf.io/

To make the test script more stable, let's use fake data in `run_python_examples.sh`. The test time for dcgan is reduced from minutes to ~5 seconds.

Test plan:
```
./run_python_examples.sh dcgan
```
image

### DIFF
--- a/run_python_examples.sh
+++ b/run_python_examples.sh
@@ -56,19 +56,7 @@ function start() {
 
 function dcgan() {
   start
-  if [ ! -d "lsun" ]; then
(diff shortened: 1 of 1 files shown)


ninginthecloud (approved):
Looks good! thanks for identifying this flaky test and making changes.

PR #1057, opened 2022-09-13
fix imagenet nondeterminism when seed is set (#1056)

### LINKED ISSUE #1056: cudnn deterministic not guaranteed when seed already set in imagenet
## Context
* Pytorch version: not related
* Operating System and version: not related

## Your Environment
* Is it a CPU or GPU environment?: GPU
* Which example are you using: `imagenet/main.py` https://github.com/pytorch/examples/blob/32f0c1e1073913e75fa75b848bccdbf011b2a569/imagenet/main.py

## Expected Behavior
The whole training will be deterministic when the seed is set, according to the annotations here:

https://github.com/pytorch/examples/blob/32f0c1e1073913e75fa75b848bccdbf011b2a569/imagenet/main.py#L91-L95

## Current Behavior

The `cudnn.benchmark = true` here will indeed introduce nondeterminism, according to [PyTorch docs](https://pytorch.org/docs/stable/notes/randomness.html#cuda-convolution-benchmarking)

https://github.com/pytorch/examples/blob/32f0c1e1073913e75fa75b848bccdbf011b2a569/imagenet/main.py#L207

I will send a pr soon.

yzs981130:
Closes #1056

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -88,6 +88,7 @@ def main():
         random.seed(args.seed)
         torch.manual_seed(args.seed)
         cudnn.deterministic = True
+        cudnn.benchmark = False
         warnings.warn('You have chosen to seed training. '
                       'This will turn on the CUDNN deterministic setting, '
                       'which can slow down your training considerably! '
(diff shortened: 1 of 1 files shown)


hudeven (approved):
Good catch! @yzs981130 thanks for the fix!

PR #1018, opened 2022-07-05
Add mps device

YuliyaPylypiv:
Added mps device to some networks

PR #1061, opened 2022-09-15
Revert "Add mps device"

msaroufim:
Reverts pytorch/examples#1018


msaroufim:
See https://github.com/pytorch/examples/issues/1060

PR #1062, opened 2022-09-18
Update .gitignore

VatsalSy:
For Mac users

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ docs/venv
 
 # development
 .vscode
+*.DS_Store
+/*.DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -19,4 +19,6 @@ docs/venv
 *~
 
 # development
-.vscode
\ No newline at end of file
+.vscode
+*.DS_Store
+/*.DS_Store
```
hudeven, line 24: @VatsalSy I guess you want to exclude "DS_Store" recursively. We can use `**/.DS_Store`
```suggestion
**/.DS_Store
```

PR #1070, opened 2022-09-27
Added mps option in generate.py

AvisP:
After training a model in Mac with mps option, when trying to run the generate script it is giving a Runtime error "Placeholder storage has not been allocated on MPS device". To avoid this issue, this change is made


hudeven (approved):
LGTM. @AvisP thanks for your contribution!

Issue #1074, opened 2022-10-01
The order of the parameters is reversed. The two parameter (save_every, total_epochs)positions should be interchanged.


RuleNHao:
https://github.com/pytorch/examples/blob/35eb8149b57a1edc13a1baa6a6376ae9a4e1d664/distributed/ddp-tutorial-series/multigpu.py#L102

RuleNHao:
It took me hours to find the bug. This is an unforgettable experience.

RuleNHao:
And it also need to use `sampler.set_epoch()` method at the beginning of each epoch.

suraj813:
@RuleNHao thanks for catching this! 

`set_epoch()` is an important suggestion to ensure random shuffling; thank you for bringing this up, I'll update the example to include this.

PR #1076, opened 2022-10-03
Add set_epoch and fix args in DDP-series example

suraj813:
Fixes issue reported in #1074  
cc: @hudeven


ananthsub (approved):

hudeven:
Thanks @suraj813 for fixing it quickly!

PR #1103, opened 2022-12-13
word language model on Jetson NX

IdiosyncraticDragon:
When running the word language model on Jetson NX, the original main.py fails caused by that torch (NVIDIA offical pytorch docker image: `l4t-pytorch:r35.1.0-pth1.11-py3`) do not have the `mps` backend. This modification has fixed the problem.

### DIFF
--- a/word_language_model/main.py
+++ b/word_language_model/main.py
@@ -58,7 +58,7 @@ torch.manual_seed(args.seed)
 if torch.cuda.is_available():
     if not args.cuda:
         print("WARNING: You have a CUDA device, so you should probably run with --cuda.")
-if ('mps' in dir(torch.backends)) and torch.backends.mps.is_available():
+if hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
     if not args.mps:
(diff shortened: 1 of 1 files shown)


word_language_model/main.py
```diff
@@ -58,7 +58,7 @@
 if torch.cuda.is_available():
     if not args.cuda:
         print("WARNING: You have a CUDA device, so you should probably run with --cuda.")
-if torch.backends.mps.is_available():
+if ('mps' in dir(torch.backends)) and torch.backends.mps.is_available():
```
hudeven, line 61: nit: I think using `hasattr` is cleaner
```suggestion
if hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
```

hudeven (commented):
@IdiosyncraticDragon Nice catch. Thanks for the fix! I left a minor comment.

IdiosyncraticDragon:
> @IdiosyncraticDragon Nice catch. Thanks for the fix! I left a minor comment.

Thank you for your suggestions~~, it is a better modification.

PR #1113, opened 2023-02-12
fix: fixed local device name in multinode example.

fjosw:
I fixed a small bug in the ddp multinode tutorial which does not allow the loading of snapshots (`AttributeError: 'Trainer' object has no attribute 'gpu_id'`). It looks like the `_load_snapshot` method was copied over from `multigpu_torchrun.py` but the attribute `gpu_id` was not renamed to `local_rank`.


soumith (approved):

soumith:
thank you!

PR #1115, opened 2023-02-19
Replace list with deque.

helpingstar:
FIX

Same PR for #1083
(There is no difference between the actor_critic.py example and reinforcement.py in using `returns`.)

The time complexity of `returns.append(0, R)` inserting at index 0 of the return list is O(N), and this can be reduced to O(1) by using a deque instead of a list.

According to the python docs:

> Deques support thread-safe, memory efficient appends and pops from either side of the deque with approximately the same O(1) performance in either direction.

> Though [list](https://docs.python.org/3/library/stdtypes.html#list) objects support similar operations, they are optimized for fast fixed-length operations and incur O(n) memory movement costs for pop(0) and insert(0, v) operations which change both the size and position of the underlying data representation.

The relevant official explanation is described on the site below.
https://docs.python.org/3/library/collections.html#collections.deque

### DIFF
--- a/reinforcement_learning/actor_critic.py
+++ b/reinforcement_learning/actor_critic.py
@@ -2,7 +2,7 @@ import argparse
 import gym
 import numpy as np
 from itertools import count
-from collections import namedtuple
+from collections import namedtuple, deque
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


msaroufim:
Curious what kind of end to end speedup you ended up seeing with this? I don't have any benchmarks running in CI

helpingstar:
sorry. I will supplement the content and PR again.

PR #1118, opened 2023-02-23
Add logic to resume ImageNet example with new learning rate

### LINKED ISSUE #816: Resume will override learning rate on command line.
If you resume from a checkpoint and want to run with a new learning rate, the saved learning rate will overwrite what you specify on the command line. The fix is simple.

kiersten-stokes:
Fixes #816 

Resume no longer overwrites learning rate specific on command line. 

(My linter also removed 2 extraneous spaces in a `dict`, I can revert this if desired 🙂)

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -215,6 +215,9 @@ def main_worker(gpu, ngpus_per_node, args):
                 best_acc1 = best_acc1.to(args.gpu)
             model.load_state_dict(checkpoint['state_dict'])
             optimizer.load_state_dict(checkpoint['optimizer'])
+            if args.lr:
+                # resume with newly specified learning rate
+                optimizer.param_groups[0]['lr'] = args.lr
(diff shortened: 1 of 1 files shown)


imagenet/main.py
```diff
@@ -215,6 +215,9 @@ def main_worker(gpu, ngpus_per_node, args):
                 best_acc1 = best_acc1.to(args.gpu)
             model.load_state_dict(checkpoint['state_dict'])
             optimizer.load_state_dict(checkpoint['optimizer'])
+            if args.lr:
```
soumith, line 218: wouldn't `if args.lr` always evaluate to true?

kiersten-stokes: shoot, yes, thank you for pointing that out! I must have been looking at `start_epoch` or something

kiersten-stokes:
@soumith just pinging again to see if this is in a good state for approval 🙂

PR #1127, opened 2023-03-19
corrected rpc

NicolasRR:
Corrected a bug on rpc batch training that prevented updating the weights manually because of a None Type error


msaroufim:
CAn you please post logs before and after?

PR #1129, opened 2023-03-29
typo fix: multi-node training command in imagenet/

Mr-Philo:
Hi there, I noticed that in imagenet/README.md, when using multi-node distributed training, the value of param '--dist-url' of node 1 should not be `IP_OF_NODE0`, but `IP_OF_NODE1`. I guess it was just a typo :)


mengwanguc:
I think the original READM (`IP_OF_NODE0`) is correct.

msaroufim:
Closing for now, would be useful to understand how the bug was detected and fixed with some logs

PR #1137, opened 2023-04-14
Notebook for training ResNet with torch.compile

agunapal:
This is an example Notebook demonstrating how to train a `ResNet` model using `torch.compile`
 - This also demonstrates some techniques to speed up training
 - Demonstrates when to use `reduce-overhead` mode


agunapal:
cc @msaroufim

msaroufim:
FYI we don't typically have notebooks in `examples` might be a better fit for tutorials, so if you'd like to remove that I can review the rest

agunapal:
@msaroufim I am passing different modes/ optimizations as optional args, build a story step-by-step. So, without the notebook, I am not sure if `main.py` would be useful as is. What do you suggest? Should I work on moving the entire thing to tutorials?

msaroufim:
Yeah tutorials might be a more natural fit

PR #1142, opened 2023-04-27
Adds torch.cuda.set_device calls to DDP examples

subramen:
The recommended best practice to avoid hangs and high memory utilization on GPU:0 is to `set_device(i)` which switches the default from GPU:0 to GPU:i

### DIFF
--- a/distributed/minGPT-ddp/mingpt/main.py
+++ b/distributed/minGPT-ddp/mingpt/main.py
@@ -8,6 +8,7 @@ from torch.distributed import init_process_group, destroy_process_group
 
 def ddp_setup():
     init_process_group(backend="nccl")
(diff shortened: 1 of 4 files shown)


rohan-varma (approved):

Neilblaze (approved):

distributed/minGPT-ddp/mingpt/main.py
```diff
@@ -8,6 +8,7 @@
 
 def ddp_setup():
     init_process_group(backend="nccl")
+    torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))
```
qiuosier, line 11: torch is not imported.

qiuosier: Also we need `import os`

PR #1149, opened 2023-05-10
Add Sequence parallel and 2D parallel examples

fduwjj:


distributed/tensor_parallelism/example.py
```diff
@@ -1,19 +1,22 @@
 import argparse
 import os
+
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
 import torch.nn as nn
 
 TP_AVAILABLE = False
 try:
```
wanchaol, line 10: I think we should remove those try catch as it's already available?

distributed/tensor_parallelism/example.py
```diff
@@ -48,17 +51,47 @@
 to use and our `parallelize_module` API will parse and parallelize the modules
 based on the given `ParallelStyle`. We are using this PyTorch native Tensor
 Parallelism APIs in this example to show users how to use them.
+
+Also one can specify the flag `run_seq_parallel` to run the example about
```
wanchaol, line 55: I think we should have a separate example for sequence parallel, maybe this file can be renamed to `tensor_parallel_example.py` where it tests TP (maybe with 2-D together), and we should have a separate example that showcase sequence parallel, so that we could have more explanation about SP, wdyt?

distributed/tensor_parallelism/example.py
```diff
@@ -74,30 +107,48 @@ def forward(self, x):
...
     """
     print(f"Running basic Megatron style TP example on rank {rank}.")
     setup(rank, args.world_size)
+
     # create a sharding plan based on the given world_size.
-    device_mesh = DeviceMesh(
-        "cuda",
-        torch.arange(args.world_size),
-    )
+    device_mesh = DeviceMesh("cuda", generate_mesh(args.world_size, args.run_2d))
```
wanchaol, line 129: Maybe we should have a separate 2-D example too? I think for examples, we should have a relatively standalone simple script without too much helper functions, so that user can easily understand how the API works

distributed/tensor_parallelism/example.py
fduwjj: We added version control in requirements. So I think we can remove this.

distributed/tensor_parallelism/example.py
fduwjj: There are so many duplicate code so I didn't split them into different files.

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -0,0 +1,83 @@
...
+from torch.distributed.tensor.parallel import parallelize_module, SequenceParallel
+from utils import cleanup, setup, ToyModel
+
+try:
+    from torch.distributed.tensor.parallel import (
+        SequenceParallel
+    )
+    SP_AVAILABLE = True
+except BaseException as e:
+    pass
```
fduwjj, line 16: This is still needed for sequence parallel for now. After 2.1 release, we can remove it.

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -0,0 +1,126 @@
...
+    # Create a optimizer for the parallelized module.
+    LR = 0.25
+    optimizer = torch.optim.SGD(model.parameters(), lr=LR)
+    # Parallelize the module based on the given Parallel Style.
+    parallel_style = SequenceParallel() if args.run_seq_parallel else PairwiseParallel()
+    model = parallelize_module(model, device_mesh, parallel_style, tp_mesh_dim=1)
+
+    # We need to register hooks for TP + FSDP integration.
+    assert (
+        enable_2d_with_fsdp()
```
wanchaol, line 84: Is this still a requirement? I thought it should automatically be registered? Can you register it when importing TP (i.e. in `tensor_parallel/__init__.py`?

wanchaol (approved):
lgtm, thanks!

distributed/tensor_parallelism/two_d_parallel_example.py
wanchaol: I mean, when user import TP, I assume they want to use 2D?

PR #1159, opened 2023-06-03
fix a typo in the FSDP example

Separius:


msaroufim:
What's the error you get before your change?

Separius:
I didn't get an error, but only two elements are being used from that tensor and the train() function, has a tensor of size 2, I thought it's just a simple typo here.

msaroufim (approved):

PR #1162, opened 2023-06-08
Image Processing with Pytorch

mechatronicsengineer95:
Hello, I would like to use it in Image Processing for calculating the similarity between two photos. I am thinking Pytorch is really well for this process. But I couldn't see any issue therefore I would like to open that issue. I would like to submit a code as an example of usage for PyTorch in this area. And I would like to know how and where can I do this.

My codes calculate the similarity between the reference image and the target image. For the data, I got some of the example pictures but the photo could change, it is necessary.


msaroufim:
Hi @mechatronicsengineer95 this might be a better fit for the tutorials repo - we have a hackathon ongoing so this is good timing

Issue #1131, opened 2023-04-10
New examples requested


msaroufim:
Hi everyone, @svekars and I are looking to increase the number of new contributions to pytorch/examples, this might be especially interesting to you if you've never contributed to an open source project before.

At a high level, we're looking for new interesting models.

So here's what you need to do
1. Check out our contributing guide: https://github.com/pytorch/examples/blob/main/CONTRIBUTING.md
2. Pick a model idea - I've listed a few below, comment on this task so others know you're working on it
3. Implement your model from scratch using PyTorch, no external dependencies will be allowed to keep the examples as educational as possible

Your implementation needs to include
1. A folder with your code which needs to define
  1. Your model architecture
  2. Training code
  4. Evaluation code 
  5. An argparser 
3. Make sure your script runs in CI so it doesn't break in the future by adding it to `run_python_examples.sh`
4. README describing any usage instructions

As an example this recent contribution by @sudomaze is a good one to follow https://github.com/pytorch/examples/pull/1003/files

Here are some model ideas

## Model ideas

* [ ] Controlnet - Guided diffusion
* [ ] NERF 
* [x] Graph Neural Network @JoseLuisC99 
* [ ] Diffusion Model, stable diffusion or any variant of the architecture you like
* [x] Vision Transformer
* [ ] Video model
* [ ] Toolformer
* [ ] Differentiable physics
* [ ] Flownet
* [ ] Dreamfusion or any 3d model
* [ ] Language Translation
* [ ] Swin transformer

But I'm quite open to anything we don't have that's cool

niyarrbarman:
Hey! I'm interested in contributing to the vision transformer model, but I don't have any prior open source contribution experience. Would it be okay for me to proceed with this project?

msaroufim:
Yes please go for it

xorsuyash:
@msaroufim I would Like to contribute to examples related to graph neural networks ,Is there any specific Dataset i should choose for this or i can choose any dataset of my own choice for examples .

msaroufim:
Yes to all of the above

aditikhare007:
@msaroufim I am interested for PyTorch Open-Source Contribution. Thank you for sharing New examples requested Note, would like to contribute to Diffusion Model, stable diffusion and Vision Transformer sections will keep you posted as my work progress.

aditikhare007:
[...]

PR #1164, opened 2023-06-11
[WIP] Adding controlnet

arunppsg:
#1131

This pull request adds a new example of controlnet model in pytorch.


msaroufim:
Hi @arunppsg are you still interested in merging this? I think you're close

arunppsg:
The original controlnet model is big and I am not able to test it in my machine. I also need to add a couple of more components to the model. I have a question: Should we make a small minimal implementation, like [min-GPT](https://github.com/pytorch/examples/tree/6fc19c76b9f5550d8d3cb38284845ac7ed14223c/distributed/minGPT-ddp) or do an implementation of full model? I think the former approach will be good for an example and latter one is already available at [ControlNet](https://github.com/lllyasviel/ControlNet).

msaroufim:
Minimal is usually better for this repo

arunppsg:
Sounds good, I will work on making a minimal prototype by end of Aug.

Issue #1175, opened 2023-07-26
`vision_transformer` `num_classes` 16 instead 10 for CIFAR10 dataset.


ninfueng:
## Context

* Pytorch version: 2.0.1+cu117
* Operating System and version: Window subsystem (Ubuntu 22.04.2 LTS)

## Your Environment

* Installed using source? [yes/no]: no
* Are you planning to deploy it using docker container? [yes/no]: no
* Is it a CPU or GPU environment?: GPU
* Which example are you using: vision_transformer
* Link to code or data to repro [if any]: [vision_transformer](https://github.com/pytorch/examples/tree/6fc19c76b9f5550d8d3cb38284845ac7ed14223c/vision_transformer)

## Expected Behavior

From there:

https://github.com/pytorch/examples/blob/6fc19c76b9f5550d8d3cb38284845ac7ed14223c/vision_transformer/main.py#L228C1-L229C88

The default `num_classes` should equal to 10 for CIFAR10 dataset.

## Current Behavior

The default `num_classes` equals to 16 instead.

## Possible Solution

Just change `num_classes` to 10.

## Steps to Reproduce

...

ninfueng:
By the way, I found an another similar problem as below:
 
https://github.com/pytorch/examples/blob/6fc19c76b9f5550d8d3cb38284845ac7ed14223c/vision_transformer/main.py#L220C1-L221C47

The default `num_heads` should be 16 (same as the `help`) or the value in `help` should be 12 (same as the default value).

ninfueng:
The fix seems to be merged in #1176, so I should close this issue. Note that the issue from the previous comment `num_heads` is still here.

PR #1176, opened 2023-07-26
Fixed default num_classes to 10 for CIFAR10

niyarrbarman:
Changed default num_classes to 10 for CIFAR10 as mentioned in #1175


niyarrbarman:
@msaroufim Can we merge this ?

PR #1201, opened 2023-11-21
[T170073014] Rewrite distributed examples for Tensor Parallel, Sequence Parallel, 2D (FSDP + TP)

lessw2020:
This PR updates the three distributed examples for Tensor Parallel, Sequence Parallel and 2D with the following main changes:
(note - internal reference - task [T170073014] Rewrite TensorParalell/SequenceParallel Examples using our new UX)

1 - move to torchrun launching (see run_.sh files) and relevant world topology introspection in the setup instead of mp.spawn. 
2 - move device mesh creation to new api, init_device_mesh
3 - use custom parallelization plans (ColwiseParallel and RowwiseParallel) rather than the previous prebuilt PairwiseParallel() and SequenceParallel()
4 - For the 2D example - used a more relevant swiglu MLP model to showcase applying 2D to a more sophisticated/llama style situation. 
5 - Adds more interactive UI for the user (start, per iter, and completion feedback).


distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+        """helper function to print only on global rank 0"""
+        if _rank==0:
+            print(f"{msg}")
+
+    print(f"Running basic Megatron style Sequence Parallel example on rank {_rank}.")
+
+    # create a device mesh based on the given world_size.
+
+    device = f"cuda"
+    device_mesh = init_device_mesh(device_type = device,mesh_shape = (_world_size,))
```
wanchaol, line 67: nit: we should just do make this simple too
```
device_mesh = init_device_mesh("cuda", (int(os.environ["WORLD_SIZE"]),))
```

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+    assert device_mesh is not None, "unable to create valid device mesh"
```
wanchaol, line 68: no need to have this assertion? iirc `init_device_mesh` would definitely give back a DeviceMesh otherwise it would error out in its init

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+
+    rank_print(f"Device Mesh created: {device_mesh=}")
```
wanchaol, line 70: hmm I feel this can be a logger.debug instead

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
 
-    # create a sharding plan based on the given world_size.
-    device_mesh = DeviceMesh("cuda", torch.arange(0, args.world_size))
 
     # create model and move it to GPU with id rank
-    model = ToyModel().cuda(rank)
+    model = ToyModel().cuda(_rank)
```
wanchaol, line 74: after `init_device_mesh`, each rank already have its corresponding GPU assigned, so we don't need to know the rank, you can simply do:

`
model = ToyModel().to("cuda")
`

lessw2020:
* test failures are related to being unable to import init_device_mesh (?), no gpu's available (?), and lastly need to modify these tests to launch via the .sh files associated with each example (to run torchscript):
~~~
  File "tensor_parallelism/tensor_parallel_example.py", line 5, in <module>
    from torch.distributed._tensor.device_mesh import init_device_mesh
ImportError: cannot import name 'init_device_mesh' from 'torch.distributed._tensor.device_mesh' (/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/torch/distributed/_tensor/device_mesh.py)
tensor parallel example failed
  File "tensor_parallelism/sequence_parallel_example.py", line 5, in <module>
    from torch.distributed._tensor.device_mesh import init_device_mesh
ImportError: cannot import name 'init_device_mesh' from 'torch.distributed._tensor.device_mesh' (/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/torch/distributed/_tensor/device_mesh.py)
sequence parallel example failed
  File "tensor_parallelism/two_d_parallel_example.py", line 18, in <module>
    from torch.distributed._tensor.device_mesh import init_device_mesh
ImportError: cannot import name 'init_device_mesh' from 'torch.distributed._tensor.device_mesh' (/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/torch/distributed/_tensor/device_mesh.py)
2D parallel example failed
Requires at least 8 GPUs to run, but got 0.
~~~

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+    )
+
+
     # Create a optimizer for the parallelized module.
-    LR = 0.25
-    optimizer = torch.optim.SGD(model.parameters(), lr=LR)
-    # Parallelize the module based on the given Parallel Style.
-    model = parallelize_module(model, device_mesh, SequenceParallel())
+    lr = 0.25
+    optimizer = torch.optim.AdamW(sp_model.parameters(), lr=lr)
```
wanchaol, line 88: nit: we can pass `foreach=True` to AdamW to make it faster

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
     # and optimizations for the sharded module.
-    for _ in range(args.iter_nums):
+    num_iters = 10
+    rank_print(f"Sequence Parallel training starting...")
+
+    for i in range(num_iters):
         # For SP, input can be different across all ranks.
-        inp = torch.rand(20, 10).cuda(rank)
-        output = model(inp)
+        inp = torch.rand(20, 10).cuda(_rank)
```
wanchaol, line 98: ditto: `inp = torch.rand(20, 10, device="cuda")` is better

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+        output = sp_model(inp)
         output.sum().backward()
         optimizer.step()
+        rank_print(f"Sequence Parallel iter {i} completed")
 
+    rank_print(f"Sequence Parallel training completed!")
```
wanchaol, line 104: logger.debug or info

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
     cleanup()
 
 
 if __name__ == "__main__":
```
wanchaol, line 108: we are running with torchrun, so I think we can remove the `__main__` too, let's make things simpler for examples ;)

distributed/tensor_parallelism/tensor_parallel_example.py
```diff
@@ -1,11 +1,17 @@
...
-import torch.multiprocessing as mp
 
-from torch.distributed._tensor import DeviceMesh
-from torch.distributed.tensor.parallel import PairwiseParallel, parallelize_module
-from utils import cleanup, setup, ToyModel
+from torch.distributed._tensor.device_mesh import init_device_mesh
+from torch.distributed._tensor import DeviceMesh, DTensor, Replicate, Shard
+
+from torch.distributed.tensor.parallel import (
+    PairwiseParallel,
```
wanchaol, line 9: ditto: we want to delete PairwiseParallel soon

distributed/tensor_parallelism/tensor_parallel_example.py
```diff
@@ -1,11 +1,17 @@
...
+    parallelize_module,
+    ColwiseParallel,
+    RowwiseParallel,
+)
+from utils import cleanup, ToyModel, torchrun_setup
```
wanchaol, line 14: this file I have exactly same comments from `sequence_parallel_example.py`, please address accordingly :)

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -2,24 +2,23 @@
 
 import torch
 import torch.distributed as dist
-import torch.multiprocessing as mp
 
 from torch.distributed._tensor import DeviceMesh
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.tensor.parallel import (
     PairwiseParallel,
```
wanchaol, line 9: same, no Pairwise

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -2,24 +2,23 @@
...
-    SP_AVAILABLE = True
-except BaseException as e:
-    pass
+
+from torch.distributed._shard.sharded_tensor import ShardedTensor
+from torch.distributed._tensor import DTensor, Replicate, sharding_prop
+from torch.distributed._tensor.device_mesh import init_device_mesh
+import os
+
+from utils import cleanup, torchrun_setup, MLP_swiglu
```
wanchaol, line 21: Let's define MLP_swiglu in each example file, it's fine to have some duplications as these are examples, we want user to see e2e workflow in a single file

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -54,74 +53,109 @@
...
+
+    # To support identical inputs for TP groups, we need the dp process group
     dp_pg = device_mesh.get_dim_groups()[0]
-    model = FSDP(model, process_group=dp_pg)
 
+    # For TP, input needs to be same across all TP ranks.
+    # while for SP, input can be different across all ranks.
+    # We will use dp_rank for setting the random seed
+    # to mimic the behavior of the dataloader.
+    dp_rank = dist.get_rank(dp_pg)
```
wanchaol, line 105: I see, this needs to be consolidate to a device mesh API, cc @wz337

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -54,74 +53,109 @@
...
-    setup(rank, args.world_size)
+    torchrun_setup()
+
+    # understand world topology
+    _rank = int(os.environ["RANK"])
+    _local_rank = int(os.environ["LOCAL_RANK"])
+    _world_size = int(os.environ["WORLD_SIZE"])
+    _local_world_size = int(os.environ["LOCAL_WORLD_SIZE"])
+
+    torch.cuda.set_device(_local_rank)
```
wanchaol, line 69: similar comments as the `sequence_parallel_example.py`, please address :)

wanchaol (commented):
First pass, I think we can do things a lot simpler. please see inline comments

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -54,74 +53,109 @@
...
-    assert (
-        enable_2d_with_fsdp()
-    ), "FSDP 2D hook is not registered. Please use PyTorch with version >= 2.0"
+    dp_size = _world_size // args.tp_size
+
+    # Create a device mesh with 2 dimensions.
+    # First dim is the data parallel dimension
+    # Second dim is the tensor parallel dimension.
+    device_mesh = init_device_mesh(device, (dp_size, args.tp_size), mesh_dim_names=("dp","tp"))
+    assert device_mesh is not None, "unable to create valid device mesh"
```
wz337, line 92: Not against having this assertion, but just curious of why we need this assertion. Have you run into scenario where device_mesh is not successfully created but it silently passed?

distributed/tensor_parallelism/run_sequence_parallel.sh
```diff
@@ -0,0 +1 @@
+torchrun --nnodes=1 --nproc_per_node=4 --rdzv_id=101 --rdzv_endpoint="localhost:5973" sequence_parallel_example.py
```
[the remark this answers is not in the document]
lessw2020: sounds good.  I've updated it to a single file:
**run_example.sh**
That takes in two optional args: 
 file_to_run.py and num_gpus
~~~
bash run_example.sh 
~~~ 
will launch twod example with 4 gpus as defaults (and print what it's launching).
User can then pass in file name of other samples and update gpus ala:
~~~
bash run_examples.sh sequence_parallel_example.py 8  
~~~
to run sp example with 8 gpus.

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,75 @@
...
+    _local_world_size = int(os.environ["LOCAL_WORLD_SIZE"])
+
+    torch.cuda.set_device(_local_rank)
+
+    def rank_print(msg):
+        """helper function to print only on global rank 0"""
+        if _rank==0:
+            print(f"{msg}")
+
+    print(f"Running basic Megatron style Sequence Parallel example on rank {_rank}.")
```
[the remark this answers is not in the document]
lessw2020: will reduce most of these, but the _rank and rank_print function is to make it cleaner to output the training info etc. to the user rather than generic show every print * every gpu and overwhelm, so do want to keep that.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: will do this on tp and sp example, but for 2D - we do some additional checks with world size so inlining it here would mean redundancy elsewhere...
example:
assert (
    _world_size % tp_size == 0
), f"World size {_world_size} needs to be divisible by TP size {tp_size}"

But will remove for those 2 that don't rely on it.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: thanks for pointing this out!

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: hmm, it doesn't work:
receive this error:
~~~
Duplicate GPU detected : rank 0 and rank 1 both on CUDA device 6000
~~~

distributed/tensor_parallelism/sequence_parallel_example.py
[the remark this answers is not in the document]
wanchaol: after `init_device_mesh`, you can get a device mesh object and call `device_mesh.get_rank()` to obtain the rank information, I would recommend not using init_process_group call anymore

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: great point! 
updated and verified all 3 examples with this change.

distributed/tensor_parallelism/sequence_parallel_example.py
wanchaol: hmmm I just did a quick try, this works seamlessly for me. NOTE: you need to delete `init_process_group` call, I think we don't want that call before init_device_mesh, otherwise you need to set your own device P887595798

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: unfortunately this results in a hang for seq parallel example and for tp example:
~~~
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:3 and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
~~~

lessw2020: for 2D though, this works so have updated there.

distributed/tensor_parallelism/tensor_parallel_example.py
lessw2020: have removed for all 3.  for reference, you don't need to mark each file re: review comments...just show on one and note that it applies to all ;)

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: this was just purely defensive programming, reflex add :) 
Per Wanchao init_device_mesh will internally assert assuming failure to make a device mesh, so have removed as it's thus a redundant check that would never be hit.

distributed/tensor_parallelism/run_sequence_parallel.sh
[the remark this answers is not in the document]
lessw2020: way nicer with single file and args - can easily control num gpus and which file to run.

distributed/tensor_parallelism/original.py
```diff
@@ -0,0 +1,127 @@
+import argparse
```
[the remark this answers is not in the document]
lessw2020: yes removed from all 3. 
I had left as was not sure if we wanted to have args passed in the future, but given not, have removed.

distributed/tensor_parallelism/sequence_parallel_example.py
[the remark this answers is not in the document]
lessw2020: oh I did not realize we can now skip the init_process_group. 
That will definitely help clean up things here.
Will update to that.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: nice - works great once I removed the init process group aspect. 
Will update all the examples.

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -1,11 +1,17 @@
...
+from torch.distributed._tensor.device_mesh import init_device_mesh
+from torch.distributed._tensor import DeviceMesh, DTensor, Replicate, Shard
+
+from torch.distributed.tensor.parallel import (
+    PairwiseParallel,
+    parallelize_module,
+    ColwiseParallel,
+    RowwiseParallel,
+)
+from utils import cleanup, ToyModel, torchrun_setup
```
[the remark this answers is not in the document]
lessw2020: Utils.py is gone.  Sample models are now all inline in the respective example files.

distributed/tensor_parallelism/tensor_parallel_example.py
lessw2020: all removed :)

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: all better.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: all updated.

distributed/tensor_parallelism/original.py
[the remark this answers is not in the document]
wanchaol: hmmm looks like this file is not deleted? what's the concern of not removing this file?

If it's about the 2-D Parallel comment block, we can copy it to two_d_parallel_example.py

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -1,25 +1,23 @@
-import argparse
```
wanchaol, line 1: nit: file name, I think let's call it fsdp_tp_example.py, `two_d_parallel_example` is too vague

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -1,25 +1,23 @@
...
-from utils import cleanup, setup, ToyModel
-try:
-    from torch.distributed.tensor.parallel import (
-        SequenceParallel
-    )
-    SP_AVAILABLE = True
-except BaseException as e:
-    pass
+
+from torch.distributed._shard.sharded_tensor import ShardedTensor
```
wanchaol, line 16: why there's `ShardedTensor` here?

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -1,25 +1,23 @@
...
+from torch.distributed._tensor import DTensor, Replicate, sharding_prop
```
wanchaol, line 17: why we need to import `sharding_prop`?

distributed/tensor_parallelism/two_d_parallel_example.py
```diff
@@ -52,76 +50,121 @@
...
+"""
+Main body of the demo of a basic version of tensor parallel by using
+PyTorch native APIs.
+"""
+tp_size = 2
+
+
+# understand world topology
+_rank = int(os.environ["RANK"])
+_local_rank = int(os.environ["LOCAL_RANK"])
```
wanchaol, line 87: why we need to have local_rank and rank defined here? 
"understand world topology", I feel this is sth why we designed a "DeviceMesh" concept, for user to understand the world topology, we should just use DeviceMesh or print DeviceMesh instead

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,69 @@
...
+
+def rank_print(msg):
+    """helper function to print only on global rank 0"""
+    if _rank==0:
+        print(f"{msg}")
+
+print(f"Running basic Megatron style Sequence Parallel example on rank {_rank}.")
+
+# create a device mesh based on the given world_size.
+_device = f"cuda"
```
wanchaol, line 70: hmm if it's just "cuda", there's no need for it to be f-string, but I'm wondering why we have this variable defined as "cuda" is pretty short and simple to inline in the code below

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,69 @@
...
+    def forward(self, x):
+        return self.out_proj(self.relu(self.in_proj(x)))
+
+
+"""
+Main body of the demo of a basic version of sequence parallel by using
+PyTorch native APIs.
+"""
+
+_rank = int(os.environ["RANK"])
```
wanchaol, line 59: Can we get the rank from DeviceMesh instead?

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -1,11 +1,17 @@
...
+    parallelize_module,
+    ColwiseParallel,
+    RowwiseParallel,
+)
 
-from torch.distributed._tensor import DeviceMesh
-from torch.distributed.tensor.parallel import parallelize_module
-from utils import cleanup, setup, ToyModel
 
 try:
```
wanchaol, line 16: also this `try` catch import block can be removed

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -33,51 +39,69 @@
...
+
+"""
+Main body of the demo of a basic version of sequence parallel by using
+PyTorch native APIs.
+"""
+
+_rank = int(os.environ["RANK"])
+
+
+def rank_print(msg):
```
wanchaol, line 62: err.. so we have this function defined on every file.. hmmm ok, I feel for this function alone it make sense to put into utils.py, we can make `rank_print` be the single function leaves in utils.py if you want to use it :)

wanchaol (commented):
Looks much better! wondering what's the reason to keep original.py and also some inline comments about imports, etc.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: everything moved to logger for all three examples:

tp_example_logging

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: everything moved to logger now.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: fstring was holdover from integrating rank_id, and _device was just to have single point of control variable for placements..but you are right, unlikely it's going to change from "cuda" atm, and likely reads better to the user. 
Will remove and inline with "cuda".

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: makes sense - updated file name and references (i.e. updated default for run_example.sh)

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: just leftover from earlier code.  Removed.
(note - can't get isort working on new dev server but that would have caught stuff like this...will try to get it working later).

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: we don't - it's leftover reference.
Removed, thanks for flagging!

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: yes, updated to use device_mesh.get_rank().

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: removed!

distributed/tensor_parallelism/two_d_parallel_example.py
lessw2020: removed.  now using mesh.get_rank() for _rank.

distributed/tensor_parallelism/sequence_parallel_example.py
lessw2020: added utils.py with rank_log function.
Updated all three example files to use.

distributed/tensor_parallelism/fsdp_tp_example.py
```diff
@@ -0,0 +1,165 @@
...
+"""
+Main body of the demo of a basic version of tensor parallel by using
+PyTorch native APIs.
+"""
+tp_size = 2
+
+logging.basicConfig(
+    format="%(asctime)s %(message)s", datefmt="%m/%d/%Y %I:%M:%S %p", level=logging.INFO
+)
+logger = logging.getLogger(__name__)
```
wanchaol, line 86: nit: maybe it make sense to also config these logging in the `utils.py` too as we defined `rank_log` inside utils.py?

distributed/tensor_parallelism/sequence_parallel_example.py
```diff
@@ -1,19 +1,23 @@
...
+from utils import rank_log
 
-try:
-    from torch.distributed.tensor.parallel import (
-        SequenceParallel
-    )
-    SP_AVAILABLE = True
-except BaseException as e:
-    pass
+logging.basicConfig(
```
wanchaol, line 17: ditto: putting these logging config to utils.py similar to rank_log

distributed/tensor_parallelism/tensor_parallel_example.py
```diff
@@ -1,11 +1,22 @@
...
+)
 
-from torch.distributed._tensor import DeviceMesh
-from torch.distributed.tensor.parallel import PairwiseParallel, parallelize_module
-from utils import cleanup, setup, ToyModel
+import logging
+from utils import rank_log
+
+logging.basicConfig(
+    format="%(asctime)s %(message)s", datefmt="%m/%d/%Y %I:%M:%S %p", level=logging.INFO
```
wanchaol, line 17: same as above

wanchaol (approved):
This looks great! Only have some nits for logging, thanks for addressing the comments!

wanchaol (approved):

distributed/tensor_parallelism/fsdp_tp_example.py
lessw2020: yes good idea.  I've moved it all over and since the utils.py is just holding logging info, updated file name to log_utils.py.  
Looks much cleaner.

distributed/tensor_parallelism/tensor_parallel_example.py
lessw2020: much cleaner now.

distributed/tensor_parallelism/fsdp_tp_example.py
```diff
@@ -46,6 +47,12 @@
 https://docs.google.com/presentation/d/17g6WqrO00rP3MsxbRENsPpjrlSkwiA_QB4r93_eB5is/
 """
 
+_min_gpu_count = 2
```
wanchaol, line 50: I think the `min_gpu_count` for FSDP + TP would be 4? at least we need 2 x 2 mesh for 2D parallel

lessw2020: great catch...was thinking of just tp when I added that. 
will update.

msaroufim (approved):

PR #1203, opened 2023-12-08
[2D] Update 2d example to use get_local_rank

wz337:
We now have a new DeviceMesh API `get_local_rank` so users would no longer need to get the underlying pg and then get the local rank from it.


lessw2020 (approved):
nice, this new api is much cleaner. 
lgtm!

PR #1215, opened 2024-01-15
Ensure the const-ness of the data member

lancerts:
## Summary
The `data` variable loads from the info.txt and encodes the info in pair <image path, ground truth label>.
The program should not modify this information during the execution and thus ensure the const-ness of the `data` variable.

## Testing Done 
Tested the program via libtorch (2.1.2) with cuda 12.1 using both the release and debug build.


lancerts:
@msaroufim FYI, rebased to main so this change is tested by the CI job and passed the test.

PR #1222, opened 2024-01-24
Improve code readability and make number of epochs a command line argument

lancerts:
## Summary

- Make number of epochs a command line argument, the CI job runs 5 epoch (compared to its default value of 30) to reduce the overall testing time
Screenshot 2024-01-24 at 2 01 36 PM

- Improve code readability by separating the discriminator creation from the main function.
- `const referece` in  `const torch::data::Example<>& batch : *data_loader` is standard (compared with the non-const version ` torch::data::Example<>& batch : *data_loader`).
- fix warning C4244: 'initializing': conversion from 'double' to 'int64_t', possible loss of data and warning C4244: 'initializing': conversion from 'double' to 'const int64_t', possible loss of data 
with change
`const int64_t batches_per_epoch =
    static_cast<int64_t>(std::ceil(dataset.size().value() / static_cast<double>(kBatchSize)));`

### DIFF
--- a/cpp/dcgan/dcgan.cpp
+++ b/cpp/dcgan/dcgan.cpp
@@ -1,5 +1,5 @@
 #include <torch/torch.h>
-
+#include <argparse/argparse.hpp>
 #include <cmath>
 #include <cstdio>
 #include <iostream>
@@ -9,9 +9,6 @@ const int64_t kNoiseSize = 100;
 
 // The batch size for training.
 const int64_t kBatchSize = 64;
-
-// The default number of epochs to train.
-int64_t kNumberOfEpochs = 30;
 
 // Where to find the MNIST dataset.
 const char* kDataFolder = "./data";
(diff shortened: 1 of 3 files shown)


cpp/dcgan/dcgan.cpp
```diff
@@ -75,7 +75,39 @@ struct DCGANGeneratorImpl : nn::Module {
...
+     nn::BatchNorm2d(256),
+     nn::LeakyReLU(nn::LeakyReLUOptions().negative_slope(0.2)),
+     // Layer 4
+     nn::Conv2d(nn::Conv2dOptions(256, 1, 3).stride(1).padding(0).bias(false)),
+     nn::Sigmoid());
+}
+
 int main(int argc, const char* argv[]) {
+
+  if (argc > 1) {
```
msaroufim, line 99: a bit ugly no? Isn't there something like argparser in the cpp community?

msaroufim (changes requested):

cpp/dcgan/dcgan.cpp
lancerts: :P yeah, it is indeed awkward. Adopted the seemingly popular [argparser](https://github.com/p-ranav/argparse) and updated PR.

msaroufim (approved):

PR #1226, opened 2024-02-01
warning added for single GPU and NCCL

Jaiaid:
* issue link referred to hint at single GPU user what to do if nccl internal check error
* nccl required version mentioned

### DIFF
--- a/imagenet/main.py
+++ b/imagenet/main.py
@@ -106,8 +106,11 @@ def main():
 
     if torch.cuda.is_available():
         ngpus_per_node = torch.cuda.device_count()
+        if ngpus_per_node == 1 and args.dist_backend == "nccl":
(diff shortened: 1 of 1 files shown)


Jaiaid:
re-committing  this change with slight modification of warning rather than assertion. Also the GPU count variable is set to one to work with later world_size variable initiation. Although I think semantically it is confusing for CPU only platform.
This is done to help users avoid some possible frustration when using default dist backend which is NCCL.

msaroufim (approved):

sebastian-burlacu:
Does this warning imply that the default backend is nccl? The '--dist-backend' argument could be documented better, in that case - as I would assume the first example given, with no --dist-backend specified, should work fine without giving a warning. As this is a warning, I am also wary of the 'requires' since that would imply it errors out if used incorrectly, but it still seems to work fine?

PR #1240, opened 2024-03-24
Language translation example added (#1131)

NoahSchiro:
Hello all, this PR seeks to add language translation to the repo as requested in issue #1131. It utilizes transformers and closely follows the Attention is All You Need paper.

I understand there is a desire to reduce dependencies for the sake of simplicity, however I had to use the spacy library. Currently, torchtext does not offer a lot of language tokenizers, just basic English. As such, we need to lean on other tokenizers, which spacy provides. The word from torchtext devs is that spacy is an "optional dependency" (source: https://github.com/pytorch/text/issues/178) so I hope you will accept it's necessity here in this example. The alternative would be to write tokenizers for every language we want to support, which I feel is antithetical to simplicity in these examples.

Other than that, this example tries to rely heavily on the tools provided in torchtext to better showcase the library as well as get users familiar with transformers and how they work!

Please let me know if there is anything you'd like to see changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ https://pytorch.org/examples/
 - [PyTorch Module Transformations using fx](./fx/README.md)
 - Distributed PyTorch examples with [Distributed Data Parallel](./distributed/ddp/README.md) and [RPC](./distributed/rpc)
 - [Several examples illustrating the C++ Frontend](cpp)
-- [Image Classification Using Forward-Forward ](./mnist_forward_forward/README.md)
+- [Image Classification Using Forward-Forward](./mnist_forward_forward/README.md)
+- [Language Translation using Transformers](./language_translation/README.md)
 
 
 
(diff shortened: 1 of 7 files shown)


msaroufim (approved):

msaroufim:
Looks like the test failed could you please take a look?

NoahSchiro:
Hi! I think the issue was that the model (with default settings) is quite large. I'm thinking the test server was potentially running out of VRAM and crashed. In any case, I have passed the appropriate flags so that it is a lot smaller.

NoahSchiro:
Thanks @msaroufim. I was going to fix that tonight after work.

In the future, how can I make sure my environment matches the testing environment so we don't have to troubleshoot here? The tests were working on my machine but seems like my environment was different.

msaroufim:
TBH I test in CI XD - but on your end you can turn https://github.com/pytorch/examples/blob/main/.github/workflows/main_python.yml into a shell script and use that, should be mostly fine

PR #1250, opened 2024-04-29
Update distributed example tests in `run_python_examples.sh`

sirutBuasai:
Update distributed example tests in `run_python_examples.sh` to use run_examples.sh.

Current script will fail distributed tests when retrieving torchrun environment variables such as [WORLD_SIZE](https://github.com/pytorch/examples/blob/d22a29f33826ff0acaeb1ceda9bcfdfff828452e/distributed/tensor_parallelism/tensor_parallel_example.py#L78) if not launched from `run_examples.sh`

### DIFF
--- a/run_python_examples.sh
+++ b/run_python_examples.sh
@@ -10,54 +10,7 @@
 # to pip install dependencies (other than pytorch), run all examples, and remove temporary/changed data files.
 # Expects pytorch, torchvision to be installed.
 
-BASE_DIR=`pwd`"/"`dirname $0`
-EXAMPLES=`echo $1 | sed -e 's/ //g'`
-
-# Redirect 'python' calls to 'python3'
-python() {
-    command python3 "$@"
-}
-
-USE_CUDA=$(python -c "import torchvision, torch; print(torch.cuda.is_available())")
-case $USE_CUDA in
-  "True")
-    echo "using cuda"
-    CUDA=1
-    CUDA_FLAG="--cuda"
-    ;;
-  "False")
-    echo "not using cuda"
-    CUDA=0
-    CUDA_FLAG=""
-    ;;
-  "")
-    exit 1;
-    ;;
-esac
-
-ERRORS=""
-
-function error() {
-  ERR=$1
-  ERRORS="$ERRORS\n$ERR"
-  echo $ERR
-}
-
-function install_deps() {
-  echo "installing requirements"
-  cat $BASE_DIR/*/requirements.txt | \
-    sort -u | \
-    # testing the installed version of torch, so don't pip install it.
-    grep -vE '^torch$' | \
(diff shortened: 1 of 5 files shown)


msaroufim:
I wonder if we should just move over those lines to https://github.com/sirutBuasai/examples/blob/release/2.3/.github/workflows/main_distributed.yaml

which calls this script https://github.com/pytorch/examples/blob/main/run_distributed_examples.sh

sirutBuasai:
Why not add these lines to the [distributed test script](https://github.com/pytorch/examples/blob/main/run_distributed_examples.sh) instead?
Any reason it is only running [fsdp_tp_example](https://github.com/pytorch/examples/blob/911816ceba94578f18433637a19f1137c3a8bb73/run_distributed_examples.sh#L62-L65)?

msaroufim:
Please go for it, it'd be an extremely useful contribution

sirutBuasai:
I've added the script to the distributed test script. I considered just calling `bash run_distributed_examples.sh` in [distributed()](https://github.com/pytorch/examples/blob/950b8d0a94549836249e32c9e114a408e2286511/run_distributed_examples.sh#L62-L68) function in the `run_python_examples.sh` so that we only declare the script entrypoint once and avoid redundancy but that might presenting some issues with error reporting since we'll be printing the error reports in both `run_distributed_examples.sh` and `run_python_examples.sh`.

Let me know what you think about whether we should consolidate the distributed tests into `run_distributed_examples.sh` or not,

msaroufim:
Indeed I'm suggesting that all distributed tests should only be in run_distributed_examples.sh

sirutBuasai:
DRY'd the code a little bit might be heavier than necessary but lmk if there are any other changes I should make.

run_python_examples.sh
```diff
@@ -10,66 +10,15 @@
...
   python main.py --dataset fake $CUDA_FLAG --mps --dry-run || error "dcgan failed"
 }
 
 function distributed() {
-    start
-    python tensor_parallelism/tensor_parallel_example.py || error "tensor parallel example failed"
-    python tensor_parallelism/sequence_parallel_example.py || error "sequence parallel example failed"
-    python tensor_parallelism/fsdp_tp_example.py || error "2D parallel example failed"
-    python ddp/main.py || error "ddp example failed"
+  bash run_distributed_examples.sh || error "distributed examples failed"
```
msaroufim, line 21: Why run these here at all? I thought the goal was to seperate out distributed vs non distributed examples

sirutBuasai: Ah I though, we wanted to still run distributed as part of run all python scripts. Just fixed.

msaroufim (approved):
Cool thank you!

PR #1261, opened 2024-05-26
Remove the unused import in MNIST example

lancerts:
Remove the unused import in MNIST example.
MNIST example succeeded in the CI


lancerts:
@msaroufim 
Not sure why the deploy failed. Cannot see the details.
WRT python examples, seem recently `language translation example failed` but doesn't seem to have related code change?

Issue #1210, opened 2024-01-04
FSDP T5 Example not working


YooSungHyun:
## Context

* Pytorch version: 3.10
* Operating System and version: ubuntu 22.04

## Your Environment

* Installed using source? [yes/no]: no
* Are you planning to deploy it using docker container? [yes/no]: no
* Is it a CPU or GPU environment?: GPU
* Which example are you using: FSDP T5 example
* Link to code or data to repro [if any]:

## Expected Behavior

training well

## Current Behavior

error raised and training stop

## Steps to Reproduce

1. launch just fsdp t5 example
2. error raised `TypeError: T5Block.forward() got an unexpected keyword argument 'offload_to_cpu'`
...

## Failure Logs [if any]

``` bash
Traceback (most recent call last):
  File "/data/bart/temp_workspace/examples/distributed/FSDP/T5_training.py", line 215, in <module>
    fsdp_main(args)
  File "/data/bart/temp_workspace/examples/distributed/FSDP/T5_training.py", line 148, in fsdp_main
    train_accuracy = train(args, model, rank, world_size, train_loader, optimizer, epoch, sampler=sampler1)
  File "/data/bart/temp_workspace/examples/distributed/FSDP/utils/train_utils.py", line 50, in train
    output = model(input_ids=batch["source_ids"],attention_mask=batch["source_mask"],labels=batch["target_ids"] )
  File "/data/bart/temp_workspace/pytorch-trainer/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/data/bart/temp_workspace/pytorch-trainer/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/data/bart/temp_workspace/pytorch-trainer/.venv/lib/python3.10/site-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 839, in forward
    output = self._fsdp_wrapped_module(*args, **kwargs)
  File "/data/bart/temp_workspace/pytorch-trainer/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/data/bart/temp_workspace/pytorch-trainer/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
[...]

PR #1273, opened 2024-06-29
Fix AC in T5 example

msaroufim:
This fixes https://github.com/pytorch/examples/issues/1210

Enabling AC is enabling compile by default which is then failing to handle the cpu offloading flag for FSDP?

This is broken on 2.2, 2.3 and nightlies

cc @awgu @anijain2305

### DIFF
--- a/distributed/FSDP/T5_training.py
+++ b/distributed/FSDP/T5_training.py
@@ -121,6 +121,7 @@ def fsdp_main(args):
         device_id=torch.cuda.current_device(),
         limit_all_gathers=fsdp_config.limit_all_gathers)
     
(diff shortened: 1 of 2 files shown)


awgu:
I think the arg signature changed for the [checkpoint wrapper](https://github.com/pytorch/pytorch/blob/main/torch/distributed/algorithms/_checkpoint/checkpoint_wrapper.py). We should remove this line:
https://github.com/pytorch/examples/blob/26de41904319c7094afc53a3ee809de47112d387/distributed/FSDP/policies/activation_checkpointing_functions.py#L16

PR #1296, opened 2024-10-24
fixed validate fn to match train fn in imagnet main file

ahmadmughees:
The transfer to the device was not consistent in the train and validate fn, so I just matched validate with train. It also reduced a couple of inconsistent calls and conditions.


msaroufim:
Superseded by some of the recent torch.accel changes

PR #1299, opened 2024-11-05
Fix python failing tests

c-p-i-o:
Summary:
Python tests are failing because of module/version incompatibilities. This change is pulling out the failing tests issue from 
https://github.com/pytorch/examples/pull/1297 into a separate commit so that the 2 changes can be committed independently.

1. Pick specific version of torchvision to fix dependency errors
2. Pin numpy to be below version 2.
3. Update Python version in python tests.

Test Plan:
Tested locally.

### DIFF
--- a/.github/workflows/main_python.yml
+++ b/.github/workflows/main_python.yml
@@ -16,11 +16,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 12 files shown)


fduwjj (approved):
Nice fix!

PR #1304, opened 2024-12-31
Fixed "--help" output in README.md

asmith512:
Updated arguments to match what main.py is looking for. Fixed incorrectly listed defaults, removed duplicate "--save-model", copied "--save_model" description from main.py.


soumith:
thanks!

PR #1311, opened 2025-01-29
Fix doc build

svekars:
python-setuptools is causing the doc build to fail: https://github.com/pytorch/examples/actions/runs/13041502252/job/36384208905
I don't think this dependency is even needed for the doc build so checking here.


dvrogozh (changes requested):
There were other issues once setuptools one is fixed. See #1331. We recently addressed that in #1333 and consider further improvements in #1336. I suggest to close this PR and continue discussion in #1336.

PR #1312, opened 2025-02-14
Bump torchvision to 0.21.0

menicosia:
When I tried this example on a fresh install (OSX Python, deployed by Brew) I got the following error:

```
% pip install -r requirements.txt
ERROR: Ignored the following yanked versions: 0.1.6, 0.1.7, 0.1.8, 0.1.9, 0.2.0, 0.2.1, 0.2.2, 0.2.2.post2, 0.2.2.post3
ERROR: Could not find a version that satisfies the requirement torchvision==0.20.0 (from versions: 0.21.0)
ERROR: No matching distribution found for torchvision==0.20.0
```

When I bumped it to v0.21.0, the error went away.

I *think* this is a harmless PR. If I'm mistaken, please reject.

### DIFF
--- a/mnist/requirements.txt
+++ b/mnist/requirements.txt
@@ -1,2 +1,2 @@
 torch
-torchvision==0.20.0
+torchvision==0.21.0


mnist/requirements.txt
```diff
@@ -1,2 +1,2 @@
 torch
-torchvision==0.20.0
+torchvision==0.21.0
```
dvrogozh, line 2: I think these examples should work on the latest torch and torchvision packages unless they deliberately test the deprecated API and soon will be dropped. As such we need to drop version pin entirely instead. That was done in the recently merged #1334 for this example. I suggest to close this PR. Otherwise, please, join discussion in issue #1329.

dvrogozh (changes requested):

PR #1315, opened 2025-02-24
Add support for Intel GPU to MNIST examples

jafraustro:
* Add support for Intel GPU to MNIST example
* Add support for Intel GPU to MNIST Forward-Forward example
* Add support for Intel GPU to MNIST using RNN example and update README with optional arguments
* Refactor argument parsing in MNIST examples. There is no need to use `default=False` with `store_true`


jafraustro:
Update in #1334

PR #1318, opened 2025-02-27
Add support for Intel GPU to Fast Neural Style example

eromomon:
Add support for Intel GPU to Fast Neural Style example


malfet (approved):
Looks ok, but would be better to use accelerator API here

atalman (approved):

PR #1323, opened 2025-03-24
Fixed install_deps so that it happens before each test run + checking for the deps that are required separately

NorseGaud:
Reasoning for the PR: The sh scripts don't do install_deps, so they fail. Each folder has a requirements file, so we should be installing the specific reqs before each test run.

Also, checks should be done to make sure torch and torchvision are installed. Pillow too as it's absence causes a failure doing the CUDA check.


msaroufim:
We've completely revamped the way we do environment isolation using uv recently

NorseGaud:
Close if you need to

PR #1324, opened 2025-04-06
Update tensor_parallel_example.py

nrothGIT:
I believe this is a small bug. If you run the current code in latest pytorch, the model will not update. I think the optimizer should be defined after the weights are parallelized as in the sequence parallelism example?


lessw2020 (approved):
good catch!  
Yes, the optimizer should be created post parallelization.
Ran the pre and post code to verify 
a) no update as claimed with the current example and 
b) proper post-parallelization optimizer application results in updated weights.

We actually expressly call out this ordering (parallelize first, then optimizer) in titan ironically: 
https://github.com/pytorch/torchtitan/blob/a4ed09c329e3f2d4c3e9f5fcbc2ce50e076f9782/torchtitan/train.py#L248

lgtm and thanks for the fix!

lessw2020:
thanks @nrothGIT for catching this and for the PR!

PR #1333, opened 2025-04-29
Update doc-build.yml and requirements for documentation build

jafraustro:
Update doc-build.yml to fix https://github.com/pytorch/examples/actions/runs/14716493656/job/41300982937 job

- Upgrade actions/checkout from v2 to v4
- Use venv to create the virtual env
- Pin sphinx version to 5.3.0 in requirements.txt with descriptions

CC: @msaroufim

Issue #1331, opened 2025-04-28
CI docs generation is broken


dvrogozh:
As can be seen in github actions ci, https://github.com/pytorch/examples/actions/workflows/doc-build.yml is broken in last few commits. That's unrelated to any actual commits in the repo and happenned due to updates on 3rd party sides (github actions runners and sphinx). First failure in (https://github.com/pytorch/examples/actions/runs/13041502252/job/36384208905):
```
E: Package 'python-setuptools' has no installation candidate
```
**Unfortunately that's just a tip of the iceberg**. Above error is `apt-get` issue and can be easily fixed with installing `python3-setuptools`. This however will uncover next issue:
```
Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```
That's https://github.com/sphinx-doc/sphinx/issues/11890. If we will change requirement to `spinx>=5`, we will further step into:
```
An error happened in rendering the page index.
Reason: UndefinedError("'style' is undefined")
```

As of now I did not identify the fix with my few trivial attempts. This needs some work.

@atalman, @msaroufim : by any chance, maybe you've met these spinx related versioning issues in other pytorch projects and can point me to where I can borrow a fix?

dvrogozh:
The following PR has addressed the build issue (thank you, @jafraustro):
* #1333

And we discuss additional improvements on top in (thank you, @soot-bit):
* #1336

PR #1336, opened 2025-04-30
This PR Improve docs build ci 

soot-bit:
This PR Improve docs build ci

**Changes:**
- removed redundant dependencies (`git`, `make`, `python3-venv`) from the GitHub Actions workflow, as they are preinstalled on `ubuntu-latest` runners. 
- Removed deprecated `panels_add_bootstrap_css`
- Created `_static` directory to resolve warnings.

**Testing:**
- Verified locally with `make html` (no errors/warnings).

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -22,7 +22,7 @@ jobs:
       run: |
         export DEBIAN_FRONTEND=noninteractive
         sudo apt-get update
-        sudo apt install -y git python3-venv make
+        sudo apt install -y git python3-venv make python3-setuptools
         echo `python3 --version`
         python3 -m venv venv 
         source venv/bin/activate 
@@ -30,6 +30,7 @@ jobs:
       id: build
     - name: Build the docset
       run: | 
+        source ../venv/bin/activate 
         cd docs
         pip install -r requirements.txt
         make html 
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -21,7 +21,6 @@ jobs:
       id: build
     - name: Build the docset
       run: | 
-        source ../venv/bin/activate 
         cd docs
         pip install -r requirements.txt
         make html 
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -18,15 +18,6 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-    - name: Dependencies
-      run: |
-        export DEBIAN_FRONTEND=noninteractive
-        sudo apt-get update
-        sudo apt install -y git python3-venv make python3-setuptools
-        echo `python3 --version`
-        python3 -m venv venv 
-        source venv/bin/activate 
-        pip install --upgrade pip setuptools 
       id: build
     - name: Build the docset
       run: | 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -65,7 +65,6 @@ html_theme_path = [pytorch_sphinx_theme.get_html_theme_path()]
 # so a file named "default.css" will overwrite the builtin "default.css".
 
 html_static_path = ["_static"]
-
 panels_add_fontawesome_latex = True
 
 html_theme_options = {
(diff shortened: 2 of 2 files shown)


.github/workflows/doc-build.yml
```diff
@@ -22,14 +22,15 @@ jobs:
       run: |
         export DEBIAN_FRONTEND=noninteractive
         sudo apt-get update
-        sudo apt install -y git python3-venv make
+        sudo apt install -y git python3-venv make python3-setuptools
```
dvrogozh, line 25: I believe docs build was fixed in the recently merged #1333. Unfortunately it did not link to the issue #1331, so you might have missed that (@jafraustro).

However, I believe you are raising few important points in this PR which might require attention.

First, I think that this entire block to install dependencies is irrelevant for the job execution. As you correctly noted, created virtual environment was not sourced in the next step. Thus, ci has worked with the system installed packages. Further, we don't need to install neither `git` or `make` on `ubuntu-latest` runner for the reason it's already preinstalled. See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

Thus, I believe we can just remove lines 21-29 from the workflow and avoid sourcing virtual environment entirely. (Virtual env is a good practice for the user work statation, but with github hosted runners you get clean environment each time when job is ran. So there is no need for additional isolation with venv.)

docs/source/conf.py
```diff
@@ -66,6 +65,7 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 
 html_static_path = ["_static"]
+html_css_files = [] # avoid undefined errors
```
dvrogozh, line 68: ci job is passing without this change after #1333 ... Is this change here actually needed?

dvrogozh (changes requested):
#1333 already addressed docs build. This PR suggests some improvements on top. I think it's better to change PR title accordingly saying for example "Improve docs build ci"

.github/workflows/doc-build.yml
```diff
@@ -18,18 +18,10 @@ jobs:
...
-        sudo apt-get update
-        sudo apt install -y git python3-venv make
-        echo `python3 --version`
-        python3 -m venv venv 
-        source venv/bin/activate 
-        pip install --upgrade pip setuptools 
       id: build
     - name: Build the docset
       run: | 
+        source ../venv/bin/activate
```
dvrogozh, line 24: should also be removed

docs/source/conf.py
```diff
@@ -66,6 +65,7 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 
 html_static_path = ["_static"]
+
```
dvrogozh, line 68: Better not to add unneeded changes.

docs/source/conf.py
```diff
@@ -41,8 +41,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ["sphinx.ext.napoleon", "sphinx.ext.autodoc", 'sphinx_panels']
-panels_add_bootstrap_css = False
```
dvrogozh, line 45: I don't have enough expertise to comment on dropping `panels_add_bootstrap_css = False`. Will leave this to other reviewers...

dvrogozh (commented):
> Created `_static` directory to resolve warnings.

You don't do that in the PR. Or that's done implicit?

> Fix documentation build: GitHub Actions, Sphinx 5.3.0, and theme comp…

I suggest to change PR title to "Improved documentation ci build"

docs/source/conf.py
soot-bit: @dvrogozh. Once again, thank you for all the guidance.
1. I’ve updated the PR title 
2. The `_static` directory was explicitly created in this PR to resolve sphinx warnings about missing paths.
image
3. sphinx-panels >=0.6.0 bundles Bootstrap CSS internally,

dvrogozh:
> The _static directory was explicitly created in this PR to resolve sphinx warnings about missing paths.

You don't have this in your PR code change. Did you forget to amend?

dvrogozh:
> its here

I overlooked it. Thank you for pointing out.

msaroufim (approved):

PR #1338, opened 2025-05-01
Use torch.accelerator API in VAE example 

eromomon:
Refactor VAE example to utilize torch.accelerator API. torch.accelerator API allows to abstract some of the accelerator specifics in the user scripts. By leveraging this API, the code becomes more adaptable to various hardware accelerators.

CC: @msaroufim, @malfet

### DIFF
--- a/vae/main.py
+++ b/vae/main.py
@@ -13,28 +13,31 @@ parser.add_argument('--batch-size', type=int, default=128, metavar='N',
                     help='input batch size for training (default: 128)')
 parser.add_argument('--epochs', type=int, default=10, metavar='N',
                     help='number of epochs to train (default: 10)')
-parser.add_argument('--no-cuda', action='store_true', default=False,
-                    help='disables CUDA training')
-parser.add_argument('--no-mps', action='store_true', default=False,
-                        help='disables macOS GPU training')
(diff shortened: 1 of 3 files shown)


msaroufim (approved):

vae/main.py
```diff
@@ -13,28 +13,31 @@
                     help='input batch size for training (default: 128)')
 parser.add_argument('--epochs', type=int, default=10, metavar='N',
                     help='number of epochs to train (default: 10)')
-parser.add_argument('--no-cuda', action='store_true', default=False,
-                    help='disables CUDA training')
-parser.add_argument('--no-mps', action='store_true', default=False,
-                        help='disables macOS GPU training')
+parser.add_argument('--accel', action='store_true',
```
dvrogozh, line 16: @eromomon : I did not see this PR before it got merged. But you've changed the default in this  example. Previously acceleration was always used (if available) and there was a flag to disable it. Now default is to not use acceleration and there is a flag to enable it. I personally like new behavior better, but **we need to change the CI script** to reflect that and that was not done:
https://github.com/pytorch/examples/blob/54e132ece5558f2b54f3f57f323bcdd71525436d/run_python_examples.sh#L148

Issue #1180, opened 2023-08-08
compile erros in cpp code 


AntonyM55:
## Context

* Pytorch version: libtorch-macos-2.0.1.zip
* Operating System and version: macos 13.4.1 (c)  

## Your Environment

* Installed using source? [yes/no]: no
* Are you planning to deploy it using docker container? [yes/no]: no
* Is it a CPU or GPU environment?: CPU
* Which example are you using: cpp/custom-dataset, cpp/distributed
* Link to code or data to repro [if any]: https://github.com/pytorch/examples/commit/741de70c4a20d9c83f811b946c186c4f83abcccb

## Current Behavior

  build custom-dataset, get 3 error 
  build distributed, get 2 error 
## Possible Solution

project custom-dataset:
error 1 and 2 because std::random_shuffle no longer exists in C++17, so I fixed std::random_shuffle with std::shuffle

e.g.
change 
  std::random_shuffle(train.begin(), train.end());
  std::random_shuffle(test.begin(), test.end());
  
to
  std::random_device rd;
  std::mt19937 seed_train(rd());
  std::mt19937 seed_test(rd());
  std::shuffle(train.begin(), train.end(), seed_train);
  std::shuffle(test.begin(), test.end(), seed_test);

project distributed:
1. change
  #include <c10d/ProcessGroupMPI.hpp>
  to 
  #include <torch/csrc/distributed/c10d/ProcessGroupMPI.hpp>

2. at line 52
auto pg = c10d::ProcessGroupMPI::createProcessGroupMPI();
return type is c10::intrusive_ptr<ProcessGroupMPI> 

and in function waitWork, the type is  
void waitWork(std::shared_ptr<c10d::ProcessGroupMPI> pg, std::vector<std::shared_ptr<c10d::ProcessGroupMPI::WorkMPI>> works)

so I changed waitWork to 
void waitWork(c10::intrusive_ptr<c10d::ProcessGroupMPI> pg, std::vector<c10::intrusive_ptr<c10d::ProcessGroupMPI::WorkMPI>> works)

and get 2 error 

/opt/libtorch/include/c10/util/intrusive_ptr.h:348:13: error: no matching function for call to 'assign_ptr_'

## Steps to Reproduce

1.
2.
...

## Failure Logs [if any]

/Users/anthony/Rnd/gits/pytorch/examples/cpp/custom-dataset/custom-dataset.cpp:90:8: error: no member named 'random_shuffle' in namespace 'std'
[...]

PR #1341, opened 2025-05-06
pass compilation; fail linking to lib/libc10d.a

davidmlw:
This change fixed the compilation errors reported in https://github.com/pytorch/examples/issues/1180 and similar the one mentioned in https://github.com/pytorch/examples/issues/1235.

But I found that the lib/libc10d.a is not found in the libtorch release https://pytorch.org/get-started/locally/ https://download.pytorch.org/libtorch/cu128/libtorch-cxx11-abi-shared-with-deps-2.7.0%2Bcu128.zip


msaroufim:
@davidmlw heads up I can't merge this unless you're comfortable signing the CLA

PR #1345, opened 2025-05-15
Add accelerate API support  for Word Language Model example

framoncg:
Refactor Word Language Model example to utilize torch.accelerator API torch.accelerator API allows to abstract some of the accelerator specifics in the user scripts. By leveraging this API, the code becomes more adaptable to various hardware accelerators.

Updated word_language_model/main.py with accelerator flag
Updated word_language_model/generate.py with accelerator flag
Updated README to match word_language_model/main.py flags
Updated run_python_examples.sh to add new accelerator flag

CC: @msaroufim, @malfet, @dvrogozh

### DIFF
--- a/word_language_model/README.md
+++ b/word_language_model/README.md
@@ -4,13 +4,13 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) or Transformer on a 
 The trained model can then be used by the generate script to generate new text.
 
 ```bash
-python main.py --cuda --epochs 6           # Train a LSTM on Wikitext-2 with CUDA.
-python main.py --cuda --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA.
-python main.py --cuda --tied               # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs.
-python main.py --cuda --epochs 6 --model Transformer --lr 5
+python main.py --accel --epochs 6           # Train a LSTM on Wikitext-2 with CUDA.
+python main.py --accel --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA.
+python main.py --accel --tied               # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs.
+python main.py --accel --epochs 6 --model Transformer --lr 5
                                            # Train a Transformer model on Wikitext-2 with CUDA.
 
-python generate.py                         # Generate samples from the default model checkpoint.
+python generate.py --accel                       # Generate samples from the default model checkpoint.
 ```
 
 The model uses the `nn.RNN` module (and its sister modules `nn.GRU` and `nn.LSTM`) or Transformer module (`nn.TransformerEncoder` and `nn.TransformerEncoderLayer`) which will automatically use the cuDNN backend if run on CUDA with cuDNN installed.
@@ -35,8 +35,7 @@ optional arguments:
   --dropout DROPOUT     dropout applied to layers (0 = no dropout)
   --tied                tie the word embedding and softmax weights
   --seed SEED           random seed
-  --cuda                use CUDA
-  --mps                 enable GPU on macOS
+  --accel               activate support for an accelerator card
   --log-interval N      report interval
   --save SAVE           path to save the final model
   --onnx-export ONNX_EXPORT
@@ -49,8 +48,8 @@ With these arguments, a variety of models can be tested.
 As an example, the following arguments produce slower but better models:
 
 ```bash
-python main.py --cuda --emsize 650 --nhid 650 --dropout 0.5 --epochs 40
-python main.py --cuda --emsize 650 --nhid 650 --dropout 0.5 --epochs 40 --tied
-python main.py --cuda --emsize 1500 --nhid 1500 --dropout 0.65 --epochs 40
-python main.py --cuda --emsize 1500 --nhid 1500 --dropout 0.65 --epochs 40 --tied
+python main.py --accel --emsize 650 --nhid 650 --dropout 0.5 --epochs 40
+python main.py --accel --emsize 650 --nhid 650 --dropout 0.5 --epochs 40 --tied
+python main.py --accel --emsize 1500 --nhid 1500 --dropout 0.65 --epochs 40
+python main.py --accel --emsize 1500 --nhid 1500 --dropout 0.65 --epochs 40 --tied
 ```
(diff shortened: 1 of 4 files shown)


word_language_model/main.py
```diff
@@ -243,11 +234,11 @@ def export_onnx(path, batch_size, seq_len):
 
 # Load the best saved model.
 with open(args.save, 'rb') as f:
-    model = torch.load(f)
+    torch.load(f, weights_only=False)
     # after load the rnn params are not a continuous chunk of memory
     # this makes them a continuous chunk, and will speed up forward pass
     # Currently, only rnn model supports flatten_parameters function.
-    if args.model in ['RNN_TANH', 'RNN_RELU', 'LSTM', 'GRU']:
+    if args.model in ['RNN_TANH', 'RNN_RELU', 'LSTM', 'GRU'] and device.type == 'cuda':
```
msaroufim, line 241: what was the error you're getting?

framoncg: Seems to be an overlook from my part. This was needed when trying a safe approach of only loading the weights but apparently it is no longer needed. I will remove it to prevent any unwanted changes

word_language_model/main.py
```diff
@@ -243,11 +234,11 @@ def export_onnx(path, batch_size, seq_len):
 
 # Load the best saved model.
 with open(args.save, 'rb') as f:
-    model = torch.load(f)
+    torch.load(f, weights_only=False)
```
dvrogozh, line 237: Can you, please, extract this change to separate PR? It also needs an update for required torch version:

https://github.com/pytorch/examples/blob/a6d10a56b63ede465ff3239a2eb5eb00136b789e/word_language_model/requirements.txt#L1

word_language_model/generate.py
```diff
@@ -21,38 +21,28 @@
...
-parser.add_argument('--cuda', action='store_true',
-                    help='use CUDA')
-parser.add_argument('--mps', action='store_true', default=False,
-                        help='enables macOS GPU training')
 parser.add_argument('--temperature', type=float, default=1.0,
                     help='temperature - higher will increase diversity')
 parser.add_argument('--log-interval', type=int, default=100,
                     help='reporting interval')
+parser.add_argument('--accel', action='store_true', default=False,
+                    help='Enables accelerated inference')
```
dvrogozh, line 29: Let's be consistent with other examples:
https://github.com/pytorch/examples/blob/a6d10a56b63ede465ff3239a2eb5eb00136b789e/fast_neural_style/neural_style/neural_style.py#L213-L214

I.e.:
```suggestion
                    help='use accelerator')
```

word_language_model/README.md
```diff
@@ -35,8 +35,7 @@ optional arguments:
   --dropout DROPOUT     dropout applied to layers (0 = no dropout)
   --tied                tie the word embedding and softmax weights
   --seed SEED           random seed
-  --cuda                use CUDA
-  --mps                 enable GPU on macOS
+  --accel               activate support for an accelerator card
```
dvrogozh, line 38: To be consistent with other examples:
https://github.com/pytorch/examples/blob/a6d10a56b63ede465ff3239a2eb5eb00136b789e/fast_neural_style/neural_style/neural_style.py#L213-L214
I.e.:
```suggestion
  --accel               use accelerator
```

word_language_model/README.md
```diff
@@ -4,13 +4,13 @@ This example trains a multi-layer RNN (Elman, GRU, or LSTM) or Transformer on a
 The trained model can then be used by the generate script to generate new text.
 
 ```bash
-python main.py --cuda --epochs 6           # Train a LSTM on Wikitext-2 with CUDA.
-python main.py --cuda --epochs 6 --tied    # Train a tied LSTM on Wikitext-2 with CUDA.
-python main.py --cuda --tied               # Train a tied LSTM on Wikitext-2 with CUDA for 40 epochs.
-python main.py --cuda --epochs 6 --model Transformer --lr 5
+python main.py --accel --epochs 6           # Train a LSTM on Wikitext-2 with CUDA.
```
dvrogozh, line 7: > `with CUDA`

I suggest to drop this from example command line and maybe add a note that example supports running on acceleration devices and list which were tried (CUDA, MPS, XPU).

dvrogozh (changes requested):

word_language_model/main.py
framoncg: If I extract the change and update the requirements to 2.7 it won't work, this change allows the example to run with the simplest code change, since leaving it as it was fails to work

jafraustro: In PyTorch 2.6, the default value for `weights_only` was set to True, and PyTorch 2.7 introduced support for the `accelerator` API.

In this pull request, we can integrate the use of the `accelerator` API in this PR.  Meanwhile, we will address the update for saving and loading models using `state_dict` in a separate pull request.

dvrogozh: > PyTorch 2.7 introduced support for the accelerator API. <...>In this pull request, we can integrate the use of the accelerator API in this PR.

From 2.6 actually. See https://docs.pytorch.org/docs/2.6/accelerator.html#module-torch.accelerator.

To integrate `torch.accelerator` we **must** update the requirement for torch to be >=2.6. Otherwise tests will simply fail. I suspect that you did not actually run the modified `run_python_examples.sh`.

> If I extract the change and update the requirements to 2.7 it won't work

I believe you are doing changes in the wrong order. First, update requirement to be able to use latest pytorch and fix issues which appear. Next, as a second step, introduce new APIs.

framoncg: I did run the modified `run_python_examples.sh` but maybe I am doing this in the wrong order. So the suggestion here is to first update requirements and fix the issues in a separate PR, close this one and create a new one for the new API?

jafraustro: First, we need to run the example with latest PyTorch and fix any issue in a separate PR.

Thanks for the feedback @dvrogozh.

dvrogozh: > the suggestion here is to first update requirements and fix the issues in a separate PR, close this one and create

Yes, but you don't need to close this PR. Just mark it as a draft while working on the update requirements PR.

dvrogozh: Here is a PR to update torch version requirement as I would do it:
* https://github.com/pytorch/examples/pull/1347

dvrogozh: @framoncg, the #1347 got merged. Please, rebase your PR.

framoncg:
hi @msaroufim I updated the flags on the CI script, tested it and it works, can you take a look?

PR #1347, opened 2025-05-16
Support torch>=2.6 in word_language_model example

dvrogozh:
Update example usage of `torch.load()` with required safe globals.

CC: @msaroufim, @framoncg, @jafraustro


word_language_model/requirements.txt
```diff
@@ -1 +1 @@
-torch<2.6
+torch>=2.6
```
dvrogozh, line 1: `torch.serialization.safe_globals` I've used in this PR appeared in torch 2.5. The `torch.load` default to load weights was changed to `True` in 2.6. Since we in any case plan to bump dependency to 2.6 by using `torch.accelerate`, I think we can update the requirement to 2.6 right away.

word_language_model/main.py
```diff
@@ -243,7 +243,31 @@ def export_onnx(path, batch_size, seq_len):
 
 # Load the best saved model.
 with open(args.save, 'rb') as f:
-    model = torch.load(f)
+    if args.model == 'Transformer':
```
dvrogozh, line 246: I think I need to add more stuff to safe_globals. Spotted this by running example manually - it's likely the difference of `--dry-run` used in CI script. Working on the fix.

dvrogozh: Was not `--dry-run` actually, but different models like `Transformer` or `GRU`. Fixed by adding more stuff to safe globals and adding few more tests to CI script.

msaroufim (approved):

PR #1348, opened 2025-05-19
Fix torch>= 2.6 generate.py compatibility in Word_language_example

framoncg:
Update example's generate.py usage of torch.load() with required safe globals.

CC: @msaroufim, @dvrogozh, @jafraustro

### DIFF
--- a/run_python_examples.sh
+++ b/run_python_examples.sh
@@ -154,10 +154,11 @@ function vision_transformer() {
 
 function word_language_model() {
   uv run main.py --epochs 1 --dry-run $CUDA_FLAG --mps || error "word_language_model failed"
+  uv run generate.py $CUDA_FLAG --mps || error "word_language_model generate failed"
   for model in "RNN_TANH" "RNN_RELU" "LSTM" "GRU" "Transformer"; do
     uv run main.py --model $model --epochs 1 --dry-run $CUDA_FLAG --mps || error "word_language_model failed"
+    uv run generate.py $CUDA_FLAG --mps || error "word_language_model generate failed"
   done
-  uv run generate.py $CUDA_FLAG --mps || error "word_language_model generate failed"
 }
 
 function gcn() {
--- a/run_python_examples.sh
+++ b/run_python_examples.sh
@@ -157,6 +157,7 @@ function word_language_model() {
   for model in "RNN_TANH" "RNN_RELU" "LSTM" "GRU" "Transformer"; do
     uv run main.py --model $model --epochs 1 --dry-run $CUDA_FLAG --mps || error "word_language_model failed"
   done
+  uv run generate.py $CUDA_FLAG --mps || error "word_language_model generate failed"
 }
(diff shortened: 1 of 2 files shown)


word_language_model/generate.py
```diff
@@ -52,7 +53,29 @@
...
+        torch.nn.modules.sparse.Embedding,
+        torch.nn.modules.transformer.TransformerEncoder,
+        torch.nn.modules.transformer.TransformerEncoderLayer,
+        RNNModel,
+        torch.nn.modules.dropout.Dropout,
+        torch.nn.modules.linear.Linear,
+        torch.nn.modules.rnn.GRU,
+        torch.nn.modules.rnn.LSTM,
+        torch.nn.modules.rnn.RNN,
+        torch.nn.modules.sparse.Embedding,
```
dvrogozh, line 75: remove duplicates, use abc order:
```suggestion
        PositionalEncoding,
        RNNModel,
        TransformerModel,
        torch.nn.functional.relu,
        torch.nn.modules.activation.MultiheadAttention,
        torch.nn.modules.container.ModuleList,
        torch.nn.modules.dropout.Dropout,
        torch.nn.modules.linear.Linear,
        torch.nn.modules.linear.NonDynamicallyQuantizableLinear,
        torch.nn.modules.normalization.LayerNorm,
        torch.nn.modules.sparse.Embedding,
        torch.nn.modules.rnn.GRU,
        torch.nn.modules.rnn.LSTM,
        torch.nn.modules.rnn.RNN,
        torch.nn.modules.transformer.TransformerEncoder,
        torch.nn.modules.transformer.TransformerEncoderLayer,
```

run_python_examples.sh
```diff
@@ -157,6 +157,7 @@ function word_language_model() {
   for model in "RNN_TANH" "RNN_RELU" "LSTM" "GRU" "Transformer"; do
     uv run main.py --model $model --epochs 1 --dry-run $CUDA_FLAG --mps || error "word_language_model failed"
   done
+  uv run generate.py $CUDA_FLAG --mps || error "word_language_model generate failed"
```
dvrogozh, line 160: If I understand correctly this is using (loading) model saved after training. If so, then safe globals will be dependent on the model used (RNN_TANH, RNN_RELU, etc.). If so, we need to run `generate.py` after each call to `main.py` above since safe globals will be different.

dvrogozh (changes requested):
Thank you for the catch. I did not realize that there is also `generate.py` script in this example which needs the same fix as we did for `main.py`.

run_python_examples.sh
framoncg: You are correct,, just pushed [78bef75](https://github.com/pytorch/examples/pull/1348/commits/78bef755bdb62b05db1107f1866ae5fd73a8fcd1) to address that part.

dvrogozh (approved):
LGTM

PR #1364, opened 2025-07-08
Update DDP example

jafraustro:
Update DDP to use the accelerator API and switch to torchrun for distributed launches

- Replace deprecated launch utility with torchrun (see PyTorch docs: https://pytorch.org/docs/stable/distributed.html#launch-utility)
- Update README to reflect torchrun usage
- Remove main.py (no referenced in documentation)
- Update CI to test example.py script


jafraustro:
> the CI is failing for Distributed examples because something cant find numpy

Hi, I changed the torch version in requirements.txt file.

 × No solution found when resolving dependencies:
  ╰─▶ Because only torch<=2.7.1 is available and you require torch>=2.8

PR #1367, opened 2025-07-08
Adding torch accelerator to FSDP2 example

dggaytan:
Updated FSDP2 train.py


jafraustro:
Hi @dggaytan,

I noticed that the FSDP2 example is not currently included in the CI (`run_distributed_examples.sh`) 
You can add it in a similar way to the following example:  -> [pytorch#1364](https://github.com/pytorch/examples/pull/1364/files#diff-ea329c10af35989ab732dc10a3dc98101f71c7fc96f1d8e186cba9088f33a216)

PR #1372, opened 2025-07-15
Migrate from FSDP1 to FSDP2 for tp_fsdp example

wwwjn:
In this diff:
- Migrated from FSDP1 to FSDP2 for tp_fsdp example


wwwjn:
cc @tianyu-l

tianyu-l (approved):

PR #1388, opened 2025-08-26
mnist: support --model-path and --load-model

Rtoax:
If specify --load-model will only run test(), if specify --model-path=A, the saved model will be renamed to A.

Example:

    python main.py --save-model --model-path=mnist_cnn.pt
    python main.py --load-model=mnist_cnn.pt


Rtoax:
@smartkiwi  Could you review this PR? thanks a lot.

PR #1389, opened 2025-08-27
mnist_forward_forward: Load dataset from comment path

Rtoax:
Load the same dataset file as other MNIST test cases.


Rtoax:
@smartkiwi Could you review this PR? thanks a lot.

PR #1393, opened 2025-09-10
Adding torch accelerator to ddp-tutorial-series example

dggaytan:
## Adding accelerator to ddp tutorials examples

- Support for multiple accelerators:
- Updated ddp_setup functions in multigpu.py, multigpu_torchrun.py, and multinode.py to use torch.accelerator for device management. The initialization of process groups now dynamically selects the backend based on the device type, with a fallback to CPU if no accelerator is available.

## Improvements to example execution:
- Added run_example.sh to simplify running tutorial examples with configurable GPU counts and node settings.
- Updated run_distributed_examples.sh to include a new function for running all DDP tutorial series examples.
## Dependency updates:
- Increased the minimum PyTorch version requirement in requirements.txt to 2.7 to ensure compatibility with the new torch.accelerator API.

CC: @msaroufim @malfet @dvrogozh


dggaytan:
@gujinghui @EikanWang

PR #1415, opened 2026-03-25
Provide an additional --xla flag to be able to run this test on TPU if available

fedebongio:
I'm a Kubernetes/Kubernetes contributor, and this is my first PR to Pytorch.

1. Trying to provide support for running this test on TPU
2. if it's a pattern that works, maybe I can generalize for other tests?
3. If I should be doing it in a different way I'm open to feedback and learning
4. I ran the test on a TPU machine and it worked, even at epoch 1:

```
fedebongio@tpu-machine-1:~/examples/mnist$ python main.py --xla --log-interval 100

WARNING:root:libtpu.so and TPU device found. Setting PJRT_DEVICE=TPU.

Train Epoch: 1 [0/60000 (0%)] Loss: 2.300604
Train Epoch: 1 [6400/60000 (11%)] Loss: 0.244346
Train Epoch: 1 [12800/60000 (21%)] Loss: 0.055622
Train Epoch: 1 [19200/60000 (32%)] Loss: 0.200639
Train Epoch: 1 [25600/60000 (43%)] Loss: 0.093245
Train Epoch: 1 [32000/60000 (53%)] Loss: 0.159420
Train Epoch: 1 [38400/60000 (64%)] Loss: 0.148803
Train Epoch: 1 [44800/60000 (75%)] Loss: 0.116827
Train Epoch: 1 [51200/60000 (85%)] Loss: 0.054931
Train Epoch: 1 [57600/60000 (96%)] Loss: 0.068974

Test set: Average loss: 0.0475, Accuracy: 9852/10000 (99%) 
```
Thanks!
Federico

### DIFF
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -144,10 +144,7 @@ def main():
     test_loader = torch.utils.data.DataLoader(dataset2, **test_kwargs)
 
     model = Net().to(device)
-    if args.xla:
-        optimizer = optim.Adam(model.parameters(), lr=1e-3)
-    else:
-        optimizer = optim.Adadelta(model.parameters(), lr=args.lr)
+    optimizer = optim.Adadelta(model.parameters(), lr=args.lr)
 
     scheduler = StepLR(optimizer, step_size=1, gamma=args.gamma)
     for epoch in range(1, args.epochs + 1):
--- a/mnist/main.py
+++ b/mnist/main.py
@@ -7,6 +7,14 @@ from torchvision import datasets, transforms
 from torch.optim.lr_scheduler import StepLR
 
 
+_XLA_AVAILABLE = False
+try:
+    import torch_xla
+    _XLA_AVAILABLE = True
+except ImportError:
+    pass
+
+
 class Net(nn.Module):
     def __init__(self):
         super(Net, self).__init__()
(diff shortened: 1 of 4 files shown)


mnist/main.py
```diff
@@ -84,48 +94,60 @@ def main():
...
                        transform=transform)
     dataset2 = datasets.MNIST('../data', train=False,
                        transform=transform)
-    train_loader = torch.utils.data.DataLoader(dataset1,**train_kwargs)
+    train_loader = torch.utils.data.DataLoader(dataset1, **train_kwargs)
     test_loader = torch.utils.data.DataLoader(dataset2, **test_kwargs)
 
     model = Net().to(device)
-    optimizer = optim.Adadelta(model.parameters(), lr=args.lr)
+    if args.xla:
```
siyuanfoundation: why does the optimizer need to change for xla?

fedebongio: Ah, great catch! I used this for debugging when was performing poorly on TPU, to then find out that the actual fix was replacing the deprecated xm.optimizer_step() with optimizer.step() + torch_xla.sync(), which correctly applies gradients in torch_xla 2.x. Adadelta works fine on XLA with the updated API. So removing this unnecesary fork. Thank you!

PR #1424, opened 2026-08-18
Zero gradients between steps in the tensor parallelism examples

HimanshuJanbandhu:
The three training loops in `distributed/tensor_parallelism` call `backward()` and `optimizer.step()` without ever calling `zero_grad()` — and none of the three files calls it anywhere:

```python
for i in range(num_iters):
    inp = torch.rand(20, 10, device=device_type)
    output = tp_model(inp)
    output.sum().backward()
    optimizer.step()
```

PyTorch accumulates into `.grad` by default, so iteration *n* steps on the sum of all gradients from iterations 1..*n* rather than on its own.

With `num_iters = 10` the loops still run and still log, so nothing looks broken — but the optimizer isn't doing what the example appears to demonstrate, and these files are a natural starting point for real training code.

Fixed in all three (`tensor_parallel_example.py`, `sequence_parallel_example.py`, `fsdp_tp_example.py`), placed after `optimizer.step()` to match the sibling `distributed/FSDP2/example.py`.

Found with [torch-preflight](https://github.com/highwaterlabs/torch-preflight).


HimanshuJanbandhu:
Gentle ping @msaroufim @H-Huang — this is a three-line change: the three `distributed/tensor_parallelism` examples call `backward()` and `optimizer.step()` in a loop with no `zero_grad()` anywhere in the file, so each iteration steps on the running sum of all previous gradients. Placed after `step()` to match the sibling `distributed/FSDP2/example.py`. CLA and CI are green.

Separately, and possibly relevant to the wider review backlog: the paths in `CODEOWNERS` are written as `./distributed/`, `./fx/`, `./github/`. GitHub matches `CODEOWNERS` patterns gitignore-style, where a leading `./` doesn't match any path — so I don't think reviewers are being auto-requested for anything in this repo. I haven't been able to verify GitHub's parse from the outside, so treat that as a question rather than a claim, but if it's right it would explain a lot and I'm happy to send a one-line fix for it separately.

